PR #1, opened 2022-05-12
Prepare for release

djc:


nrempel (commented):
This looks really good! The example in the README is pretty long. Do you think we should move it to the examples directory to be run?

nrempel (approved):

djc:
> This looks really good! The example in the README is pretty long. Do you think we should move it to the examples directory to be run?

I started out that way, but there's nothing you can actually run, since J. Random User doesn't usually have trivial control over a DNS server to handle a challenge. So then dealing with all the dev-dependencies becomes a bit of a pain for something that, in the end, doesn't actually work.

The long example is definitely not ideal but at least provides any easy jumping off point for implementing your own/seeing how stuff fits together.

PR #7, opened 2022-11-13
tls_value added to KeyAuthorization for TLS-ALPN-01

SajjadPourali:
Hey,

To create TLS-ALPN-01 need call ring internally to hash the key authorization. So the existence of this function can make it easier, similar to the DNS one.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -102,8 +102,9 @@ impl std::error::Error for Problem {}
 
 /// The response value to use for challenge responses
 ///
-/// Use [`KeyAuthorization::dns_value()`] for DNS challenges and
-/// [`KeyAuthorization::as_str()`] for other challenge types.
+/// Use [`KeyAuthorization::dns_value()`] for DNS challenges,
+/// [`KeyAuthorization::tls_value()`] for TLS challenges, and
+/// [`KeyAuthorization::as_str()`] for HTTP challenges.
 ///
 /// <https://datatracker.ietf.org/doc/html/rfc8555#section-8.1>
 pub struct KeyAuthorization(pub(crate) String);
@@ -117,6 +118,11 @@ impl KeyAuthorization {
     /// Get the base64-encoded SHA256 digest of the key authorization
     pub fn dns_value(&self) -> String {
         base64::encode_config(digest(&SHA256, self.0.as_bytes()), URL_SAFE_NO_PAD)
+    }
+    
+    /// Get the SHA256 digest of the key authorization
+    pub fn tls_value(&self) -> Vec<u8> {
+        digest(&SHA256, self.0.as_bytes()).as_ref().to_vec()
     }
 }
 


src/types.rs
```diff
@@ -118,6 +119,11 @@ impl KeyAuthorization {
     pub fn dns_value(&self) -> String {
         base64::encode_config(digest(&SHA256, self.0.as_bytes()), URL_SAFE_NO_PAD)
     }
+    
+    /// Get the SHA256 digest of the key authorization
+    pub fn tls_value(&self) -> Vec<u8> {
```
djc, line 124: IIRC the value here is the same for ALPN and HTTP, right? So maybe it should just be called `value`? Also I think we could have this return the `digest()` directly as an `impl AsRef<[u8]>`, avoiding the extra allocation.

djc (commented):
Hey, thanks for the PR! Would be great to merge something like this, see some suggestions below.

src/types.rs
SajjadPourali: https://www.rfc-editor.org/rfc/rfc8737.txt (end of sec.3), so it should be the result of the SHA-256.
I am not sure how it is possible to return the `digest()` value directly as `AsRef<[u8]>` without unsafe code because it will be considered a local variable, and it is impossible to return its reference. I would appreciate it if you give me a clue to do it.

djc: `digest()` returns a `Digest` which implements `AsRef<[u8]>`, so you can just return the `Digest` directly (but I don't want it to appear in the type signature directly to prevent coupling with the *ring* API).

SajjadPourali: > `digest()` returns a `Digest` which implements `AsRef<[u8]>`, so you can just return the `Digest` directly (but I don't want it to appear in the type signature directly to prevent coupling with the _ring_ API).

Gotcha, I thought you were talking about wrapping the `as_ref`. What do you think about implementing it `as_bytes`? Then we internally call it for the DNS one.

djc: I think the `as_` prefix is usually used for accessors that don't do much actual work, so I don't that name is a good fit here  - could be `to_bytes()` though, and using it inside `dns_value()` makes sense to me!

djc:
Should still yield the `impl AsRef<[u8]>` as explained previously.

src/types.rs
```diff
@@ -116,10 +117,10 @@ impl KeyAuthorization {
     }
 
     /// Get the SHA256 digest of the key authorization
-    pub fn to_bytes(&self) -> Vec<u8> {
-        digest(&SHA256, self.0.as_bytes()).as_ref().to_vec()
+    pub fn to_bytes(&self) -> impl AsRef<[u8]> {
```
SajjadPourali, line 120: @djc : like this?

src/types.rs
```diff
@@ -121,7 +121,7 @@ impl KeyAuthorization {
     pub fn to_bytes(&self) -> impl AsRef<[u8]> {
         digest(&SHA256, self.0.as_bytes())
     }
-    
+
     /// Get the base64-encoded SHA256 digest of the key authorization
```
SajjadPourali, line 125: For the lint checker error

djc:
I ended up renaming `to_bytes()` as `digest()` and rewrote some of the docs and published the result as 0.1.1 on crates.io.

PR #8, opened 2022-11-30
Add certificate provisioning example

nrempel:
This PR pulls the code in the README into an example that can be run with:

```
cargo run --example provision -- --name sub.domain.com
```

The process try to provision a certificate on the Let's Encrypt staging environment and will wait for the user to update their DNS record to solve the challenge before proceeding.


Cargo.toml
```diff
@@ -18,3 +18,11 @@ ring = { version = "0.16.20", features = ["std"] }
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.78"
 thiserror = "1.0.30"
+
+[dev-dependencies]
+anyhow = "1.0.66"
+clap = { version = "4.0.29", features = ["derive"] }
```
djc, line 24: Looks like this will need an MSRV bump (we should also set the `rust-version` in the `package` metadata if it's > 1.56).

examples/provision.rs
```diff
@@ -0,0 +1,144 @@
...
+    info!("private key:\n\n{}", cert.serialize_private_key_pem());
+    info!(
+        "account credentials:\n\n{}",
+        serde_json::to_string_pretty(&account.credentials()).unwrap()
+    );
+
+    Ok(())
+}
+
+/// SMTP server
```
djc, line 139: Probably remove/change this comment?

djc (commented):
Makes sense! Maybe we should remove all the code from the README in favor of linking to this file?

djc (approved):

examples/provision.rs
nrempel: Whoops.

Cargo.toml
nrempel: Since these are dev-dependencies only used for the example, can we leave MSRV as-is?

djc: Makes CI a little tricky. Plus, the new MSRV is old enough at this point that it probably doesn't matter. The majority of the ecosystem sticks pretty close to stable.

nrempel: Bumped edition and rust-version

PR #11, opened 2023-01-27
Apply clippy suggestions for 1.67

djc:


valkum (commented):
The bump of the MSRV (inline strings in 1.58) should be reflected in the README as well, I guess.

djc:
Updated MSRV notice in the README, thanks!

valkum (approved):

PR #12, opened 2023-01-30
Remove hiring notice from README

djc:


djc:
MSRV fail will be fixed with #11.

valkum (approved):

PR #15, opened 2023-03-15
Decouple finalization from retrieving the request

### LINKED ISSUE #14: Support to get the certificate URL only after order finalization status is valid
Hey, run into some issue with the certificates URL retrieval, after an order finalization. I believe it would be helpful for `instant-acme` Order struct to have an implementation for `poll_and_finalize`, as in: https://github.com/certbot/certbot/blob/7a6752a68ed77e73c2b29ab20d3ca8927f4fa7b0/acme/acme/client.py#L240 when finalizing an order, upon checking the certificate URL. This is needed because of https://github.com/letsencrypt/pebble/issues/402 basically. If this makes sense, I would be happy to provide support for it.

djc:
This should not require any extra requests in the case the order is `valid` immediately, as we've seen in the past, but allows the caller to check multiple times if the order takes more time to become valid.

Fixes #14.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,7 +36,8 @@ use types::{
 pub struct Order {
     account: Arc<AccountInner>,
     nonce: Option<String>,
-    order_url: String,
+    url: String,
+    state: OrderState,
 }
 
 impl Order {
@@ -55,12 +56,9 @@ impl Order {
     /// After the challenges have been set up, check the [`Order::state()`] to see
     /// if the order is ready to be finalized (or becomes invalid). Once it is
     /// ready, call `Order::finalize()` to get the certificate.
-    pub async fn authorizations(
-        &mut self,
-        authz_urls: &[String],
-    ) -> Result<Vec<Authorization>, Error> {
-        let mut authorizations = Vec::with_capacity(authz_urls.len());
-        for url in authz_urls {
+    pub async fn authorizations(&mut self) -> Result<Vec<Authorization>, Error> {
+        let mut authorizations = Vec::with_capacity(self.state.authorizations.len());
+        for url in &self.state.authorizations {
             authorizations.push(self.account.get(&mut self.nonce, url).await?);
         }
         Ok(authorizations)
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -129,9 +127,28 @@ impl Order {
         self.account.get(&mut self.nonce, challenge_url).await
     }
 
-    /// Get the current state of the order
-    pub async fn state(&mut self) -> Result<OrderState, Error> {
-        self.account.get(&mut self.nonce, &self.order_url).await
+    /// Refresh the current state of the order
+    pub async fn refresh(&mut self) -> Result<&OrderState, Error> {
```
djc, line 158: Not sure about this name. I also tried `update()` but I think I like `refresh()` better.

src/lib.rs
```diff
@@ -129,9 +154,28 @@ impl Order {
...
+        let rsp = self
+            .account
+            .post(None::<&Empty>, self.nonce.take(), &self.url)
+            .await?;
+
+        self.nonce = nonce_from_response(&rsp);
+        self.state = Problem::check::<OrderState>(rsp).await?;
+        Ok(&self.state)
+    }
```
valkum, line 167: Can't we let `certificate` take a timeout value, and then let it try to refresh the order until the order is valid or invalid, or the timeout triggers? It seems like a nicer API. 
The user can trigger another round of trying to get the cert by calling `certificate` again. 
I also found nothing about rate limits for getting the current order state. Let's Encrypt and the ACME spec only mention rate limits for creation of orders.

On the other hand, maybe this would be something for a high-level API (maybe in another create) which hides all the internals (states, etc).

I like the split into `finalize` and `certificate` as it better reflects what is going on spec-wise.

djc: I kind of intentionally wanted to keep instant-acme pretty opinionated. Providing a backoff strategy triggered sort of runs counter to that idea, so I'd prefer to avoid that. (Also, in practice I'm not sure it's possible in practice to observe the processing state on Let's Encrypt staging or production infra, so I kinda suspect this is mostly an issue when you run pebble in an underpowered environment?)

valkum (approved):

PR #18, opened 2023-04-23
Annotate Api::Problem error with thiserror::transparent

dvdsk:
Error variant `Api::Problem` prints "API error: {problem}" however the Problem struct has display impl that starts by printing "API error". Together this results in "API error: API error" followed by the details.

This fixes that by annotating `Api::Problem` with `thiserror`'s `transparent` attribute which forwards the display impl to the underlying types display impl.


djc:
Nice, thanks!

PR #19, opened 2023-04-23
Add comment to prevent swapping fields causing wrong early return 

dvdsk:
See PR #17 for the original issue

edit: worked on the wrong branch and then pressed PR so it took a rebase to get this to merge cleanly. Which is why the force-push is there

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -246,7 +246,7 @@ impl Account {
         Ok(Order {
             account: self.inner.clone(),
             nonce,
-            // Order of fields matters! We return errors from Problem::check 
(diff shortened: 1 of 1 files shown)


djc (approved):
Sorry for the quick merge on the last one, should have waited for this! Thanks for fixing it up, can you fix the formatting?

dvdsk:
linter is crashing with: 
```
thread 'main' panicked at 'installed manifest should have a known target', src/toolchain.rs:931:24
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Locally `cargo fmt --all -- --check` runs without issue however. I think its fine to merge.

PR #21, opened 2023-05-15
Add support for EAB in newAccount requests

picoHz:
Some issuers require an additional field, called EAB (ExternalAccountBinding), in a newAccount request. 
I've tested this code with Google Trust Services using the HTTP challenge.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,6 @@ use hyper::client::HttpConnector;
 use hyper::header::{CONTENT_TYPE, LOCATION};
 use hyper::{Body, Method, Request, Response};
 use ring::digest::{digest, SHA256};
-use ring::hmac;
 use ring::rand::SystemRandom;
 use ring::signature::{EcdsaKeyPair, ECDSA_P256_SHA256_FIXED_SIGNING};
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -458,6 +462,43 @@ trait Signer {
...
+            external_account_binding: if let Some(eab) = &account.external_account_binding {
+                Some(eab.sign(key, server_url)?)
+            } else {
+                None
+            },
+        })
+    }
+}
+
+impl<'a> ExternalAccountBinding<'a> {
```
djc, line 480: Style nit: this should go right underneath the defition of `ExternalAccountBinding`.

src/types.rs
```diff
@@ -349,6 +371,8 @@ impl LetsEncrypt {
 pub(crate) enum SigningAlgorithm {
     /// ECDSA using P-256 and SHA-256
     Es256,
+    /// HMAC with SHA-256,
```
djc, line 374: Given that RFC 8555 maintains a somewhat strict separation between MAC-based algorithms and other algorithms, I'd prefer a bit of stricter separation here. I'd propose an `enum JwsAlgorithm { Mac(MacAlgorithm), Signature(SignatureAlgorithm) }` for use in `Header`. Maybe then we could have `MacAlgorithm` and `SignatureAlgorithm` implement and have `Jwk` take a type implementing that trait as an argument?

src/lib.rs
```diff
@@ -458,6 +462,43 @@ trait Signer {
     fn key(&self) -> &Key;
 }
 
+impl<'a> JoseNewAccount<'a> {
```
djc, line 465: Style: this should go underneath the defintion of `JoseNewAccount` (if we keep that).

src/types.rs
```diff
@@ -244,6 +244,28 @@ pub struct NewAccount<'a> {
...
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct JoseNewAccount<'a> {
+    pub(crate) contact: &'a [&'a str],
+    pub(crate) terms_of_service_agreed: bool,
+    pub(crate) only_return_existing: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) external_account_binding: Option<JoseJson>,
```
djc, line 268: I don't really like the duplication of the `NewAccount` fields here. Do you think we could do this by creating a custom `Serialize` implementation for `ExternalAccountBinding` instead?

djc (commented):
Thanks for working on this! Do you think this merits a mention in the feature list in the README, as well?

src/types.rs
picoHz: I think it's not that simple because we can't pass the necessary parameters for signing `ExternalAccountBinding` to the custom `Serialize` implementation.

picoHz: Here is another possible solution:

```
#[derive(Debug, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct NewAccount<'a> {
    pub contact: &'a [&'a str],
    pub terms_of_service_agreed: bool,
    pub only_return_existing: bool,
    #[serde(skip)]
    pub external_account_binding: Option<ExternalAccountBinding<'a>>,
}

#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
pub(crate) struct NewAccountWithSignedEab<'a> {
    #[serde(flatten)]
    pub(crate) new_account: NewAccount<'a>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub(crate) external_account_binding: Option<JoseJson>,
}
```

djc:
Merged a variant of this as #22.

PR #22, opened 2023-05-17
Support external account binding

djc:
@picoHz here's an experiment to support external account binding with an API that's a bit cleaner, IMO (avoiding code duplication). Can you review it and, ideally, test that it works?

(Alternative to #21.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,7 +23,8 @@ pub use types::{
     Identifier, LetsEncrypt, NewAccount, NewOrder, OrderState, OrderStatus, Problem,
 };
 use types::{
-    DirectoryUrls, Empty, FinalizeRequest, Header, Jwk, KeyOrKeyId, Signer, SigningAlgorithm,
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -203,11 +205,28 @@ impl Account {
...
+    ) -> Result<Account, Error> {
         let client = Client::new(server_url).await?;
         let key = Key::generate()?;
+        let payload = NewAccountPayload {
+            new_account: account,
+            external_account_binding: external_account
+                .map(|eak| {
+                    JoseJson::new(
+                        Some(&Jwk::new(&key.inner)),
+                        eak.header(None, server_url),
```
picoHz, line 221:
```suggestion
                        eak.header(None, &client.urls.new_account),
```

picoHz:
@djc It worked for Google Domains and ZeroSSL 👍

djc:
Thanks for the feedback and testing! (And also, thanks for maintaining the mediatype crate, which we use.)

This just got published as 0.3 on crates.io.

PR #24, opened 2023-05-30
Abstract over different HttpClient implementations

### LINKED ISSUE #23: Support for user-injected hyper client
Hi!

What are your thoughts on supporting a user-injected hyper client? I'm referring to the client used [here](https://github.com/instant-labs/instant-acme/blob/main/src/lib.rs#L350).

djc:
Fixes #23.


rikonor:
Wow! Thanks so much for going to the trouble, @djc.

rikonor (approved):

src/lib.rs
```diff
@@ -533,15 +593,36 @@ fn nonce_from_response(rsp: &Response<Body>) -> Option<String> {
...
+                    .enable_http2()
+                    .build(),
+            ),
+        )
+    }
+}
 
-    hyper::Client::builder().build(https)
+/// A HTTP client based on [`hyper::Client`]
+pub trait HttpClient {
```
Daniel-Bloom-dfinity, line 623: Can you add an `impl` for all `Client`s?

```rust
impl<C> HttpClient for hyper::Client<C> where
    C: hyper::client::connect::Connect + Clone + Send + Sync + 'static {
    fn request(&self, req: Request<Body>) -> ResponseFuture {
        <hyper::Client<C>>::request(self, req)
    }
}
```

djc: Good idea, done.

djc:
Published as 0.3.1.

Issue #25, opened 2023-06-01
Client is not Send


surban:
When updating to instant-acme 0.3 I get the following error:

```
note: future is not `Send` as this value is used across an await
   --> /home/surban/.cargo/registry/src/index.crates.io-6f17d22bba15001f/instant-acme-0.3.1/src/lib.rs:236:69
    |
236 |             Client::new(server_url, Box::<DefaultClient>::default()).await?,
    |                                     ------------------------------- ^^^^^^ await occurs here, with `Box::<DefaultClient>::default()` maybe used later
    |                                     |
    |                                     has type `Box<dyn HttpClient>` which is not `Send`
...
239 |     }
    |     - `Box::<DefaultClient>::default()` is later dropped here
```

This makes it impossible to use instant-acme on a task spawned via `tokio::spawn`.

Version 0.2 did not have this limitation.

Could you either suggest a workaround or change the contained `dyn HttpClient` to `dyn HttpClient + Send + Sync` ?

PR #26, opened 2023-06-01
Re-Add Send,Sync to HttpClient with option to remove

dvdsk:
Adresses https://github.com/instant-labs/instant-acme/issues/25 by adding Send and Sync bounds to the HttpClient trait. Adds feature "single-thread" that removes those bounds again.

Rust specifies features should be addative (enable) functionality instead of restricting it. As adding bounds would restrict trait implementation I chose to let the feature remove the bounds.

To maintain semver the minor version needs to be increased.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ rust-version = "1.60.0"
 
 [features]
 default = ["hyper-rustls"]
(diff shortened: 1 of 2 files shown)


dvdsk:
closed in favor of #27

PR #27, opened 2023-06-02
Add some bounds for HttpClient trait

### LINKED ISSUE #25: Client is not Send

djc:
@surban @dvdsk I think I'd prefer to keep this simple -- it seems very unlikely to me that anyone started using instant-acme 0.3.1 with a non-`Send + Sync + 'static` client in the past 2 days. Does this work for your use cases?

Fixes #25. Alternative to #26.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "instant-acme"
-version = "0.3.1"
+version = "0.3.2"
 edition = "2021"
 license = "Apache-2.0"
 description = "Async pure-Rust ACME client"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -621,7 +621,7 @@ impl Default for DefaultClient {
 }
 
 /// A HTTP client based on [`hyper::Client`]
-pub trait HttpClient {
+pub trait HttpClient: Send + Sync + 'static {
     /// Send the given request and return the response
     fn request(&self, req: Request<Body>) -> ResponseFuture;
 }


dvdsk:
Works for me. My pr was mostly about presenting the option in case you needed a non send-sync client internally. If you do not then this is a far better solution.

dvdsk:
I think we need to yank 0.3.1 too, since the bounds missing is now a regression.

djc:
Going to merge this to unblock our own usage of this. @surban please let me know if you have any feedback!

surban:
0.3.2 is working perfectly for me. Thanks!

PR #28, opened 2023-06-12
Replace `ResponseFuture` with `Box<dyn>`

Daniel-Bloom-dfinity:
Re: [this](https://github.com/instant-labs/instant-acme/issues/23#issuecomment-1586801208), turns out adding an associated type to `HttpClient` would break the trait objectivity. So instead, more `Box<dyn>`!


djc:
But let's do without the `Box::into_pin()`, please...

Daniel-Bloom-dfinity:
With a little refactoring, yes.

PR #34, opened 2023-08-01
Improve credential handling

djc:
Only yield `AccountCredentials` (which gives access to the private key) at `Account` creation time. Store the `directory` in the `AccountCredentials` and prefer using it to acquire new `DirectoryUrls` if available at deserialization time.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 use std::fmt;
 
 use base64::prelude::{Engine, BASE64_URL_SAFE_NO_PAD};
@@ -53,10 +52,11 @@ impl From<&'static str> for Error {
 /// the account credentials to a file or secret manager and restore the
 /// account from persistent storage.
 #[derive(Deserialize, Serialize)]
-pub struct AccountCredentials<'a> {
-    pub(crate) id: Cow<'a, str>,
+pub struct AccountCredentials {
+    pub(crate) id: String,
     pub(crate) key_pkcs8: String,
-    pub(crate) urls: Cow<'a, DirectoryUrls>,
+    pub(crate) directory: Option<String>,
+    pub(crate) urls: Option<DirectoryUrls>,
 }
 
 /// An RFC 7807 problem document as returned by the ACME server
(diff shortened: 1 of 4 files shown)


src/types.rs
```diff
@@ -54,11 +54,51 @@ impl From<&'static str> for Error {
...
+    use std::fmt;
+
+    use base64::prelude::{Engine, BASE64_URL_SAFE_NO_PAD};
+    use serde::{de, Deserializer, Serializer};
+
+    pub(crate) fn serialize<S>(key_pkcs8: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let encoded = BASE64_URL_SAFE_NO_PAD.encode(key_pkcs8.as_ref());
```
valkum, line 74: Isn't this basically doing PEM? https://www.rfc-editor.org/rfc/rfc7468#page-13
I think it makes sense to go full PEM and add the `BEGIN` and `END` parts here as well. This should allow more interop.

valkum (approved):
Looks good. Just some small remarks for some improvements.

src/types.rs
djc: I thought about it. (a) I don't really want to invest in writing a robust PEM parser, and I kind of don't want to add another dependency for it, (b) if we support PEM I worry that users will expect us to handle a bunch of other kinds of secret keys which PEM supports but which I'd prefer to avoid supporting.

src/types.rs
```diff
@@ -54,11 +54,51 @@ impl From<&'static str> for Error {
 #[derive(Deserialize, Serialize)]
 pub struct AccountCredentials {
     pub(crate) id: String,
-    pub(crate) key_pkcs8: String,
+    /// Stored in DER, serialized as base64
+    #[serde(with = "pkcs8_serde")]
+    pub(crate) key_pkcs8: Vec<u8>,
```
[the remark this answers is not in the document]
djc: I agree that `key_pkcs8` is technically a pleonasm. However, in code, I think the context of how the key is encoded is useful, for example in code like `Key::from_pkcs8_der(credentials.key_pkcs8.as_ref())?` I find it quite helpful to see that the function name expectation matches the data I'm passing in, since expressing these kinds of contraints in the type system (like `Pkcs8(Der(private_key))`) gets a bit cumbersome and is probably overkill in this case.

djc:
Thanks for reviewing, good comments! I think overall this hits a good balance, so going to merge this for now.

PR #38, opened 2024-01-24
Add certificate revocation

Moudoux:
Adds certificate revocation, though I'm not entirely sure if the impl should be located in the types.rs.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -4,6 +4,7 @@ use base64::prelude::{Engine, BASE64_URL_SAFE_NO_PAD};
 use hyper::{Body, Response};
 use ring::digest::{digest, Digest, SHA256};
 use ring::signature::{EcdsaKeyPair, KeyPair};
+use rustls_pki_types::CertificateDer;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -271,6 +271,52 @@ pub struct NewOrder<'a> {
...
+    /// The certificate should not contain a header or footer or line breaks
+    pub certificate: String,
+    /// Reason for revocation
+    pub reason: Option<i32>,
+}
+
+impl CertificateRevocation {
+    /// Create a new revocation request from a certificate.
+    /// The certificate cannot be a chain.
+    pub fn new(certificate: &str, reason: Option<RevocationReason>) -> CertificateRevocation {
```
djc, line 304: If we're going to do a bunch of processing here on the `certificate` `&str`, it should not be a `pub` field. Instead of having this constructor on `CertificateRevocation` (and making the fields private), I would suggest that we add a dependency on rustls-pki-types, take a `CertificateDer` (which can decode from PEM using the rustls-pemfile crate -- which can be handled by the caller) and then base64url encode it during serialization with another custom `Serialize` impl. What do you think?

djc (commented):
Thanks for adding this! Have some thoughts on how to expose this in the API.

Moudoux:
I've done most of the changes. The `RevocationRequest::new` method  now accepts a `CertificateDer` which is encoded to base64 url safe in the new method. It is not possible to implement a foreign trait for a foreign struct afaik.

djc (approved):
LGTM, have you tested that this works?

djc:
I've released this as 0.4.2.

PR #43, opened 2024-01-31
Be more lenient with challenge error field

h33p:
Zerossl appears to have started to include an empty object within the `error` field of the challenges, like so:

```
"challenges":[
  {
    "type":"http-01",
    "url":"https://acme.zerossl.com/v2/DV
90/<...>",
    "status":"invalid",
    "error":{},
    "token":"<...>"
  }
]
```

Not sure if they are behaving within spec, but failing to filter out empty body leads to instant-acme breaking on such 200 response:

```
Json(Error("missing field `type`", line: 1, column: 245))
```

I changed it so that the error field defaults to `None`, if it fails to deserialize. Ideally, you'd want to check for empty `serde_json::Value`, but alas, deserializer error types are incompatible when further deserializing `serde_json::Value`.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -7,7 +7,8 @@ use ring::signature::{EcdsaKeyPair, KeyPair};
 use rustls_pki_types::CertificateDer;
 use serde::de::DeserializeOwned;
 use serde::ser::SerializeMap;
-use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


djc:
Thanks for the feedback! I think this is actually legal, and as such I've proposed an alternative fix in #44. Please give me your feedback in that PR.

PR #50, opened 2024-06-23
Update to hyper dependency to version 1

kellpossible:


src/types.rs
```diff
@@ -30,6 +33,12 @@ pub enum Error {
     /// HTTP request failure
     #[error("HTTP request failure: {0}")]
     Http(#[from] hyper::Error),
+    /// HTTP request failure
+    #[error("HTTP request failure: {0}")]
+    HttpHttp(#[from] hyper::http::Error),
```
kellpossible, line 38: Not really sure what the best way to handle this was

Cargo.toml
```diff
@@ -27,7 +29,7 @@ thiserror = "1.0.30"
 [dev-dependencies]
 anyhow = "1.0.66"
 clap = { version = "4.0.29", features = ["derive"] }
-rcgen = "0.12"
+rcgen = "0.12.1"
```
djc, line 32: This is unnecessary -- Cargo uses `^` for dependencies implicitly.

Cargo.toml
```diff
@@ -15,9 +15,11 @@ categories = ["web-programming", "api-bindings"]
 default = ["hyper-rustls"]
 
 [dependencies]
-base64 = "0.21.0"
-hyper = { version = "0.14.18", features = ["client", "http1", "http2"] }
-hyper-rustls = { version = "0.24", default-features = false, features = ["http1", "http2", "native-tokio", "tls12"], optional = true }
+base64 = "0.22.1"
```
djc, line 18: I'd like to have an upgrade to base64 `0.22` in a separate commit or PR.

src/lib.rs
```diff
@@ -448,16 +455,20 @@ struct Client {
 }
 
 impl Client {
-    async fn new(server_url: &str, http: Box<dyn HttpClient>) -> Result<Self, Error> {
+    async fn new(directory_url: &str, http: Box<dyn HttpClient>) -> Result<Self, Error> {
```
djc, line 458: This unrelated rename should be separately motivated and in a separate commit/PR.

src/lib.rs
```diff
@@ -134,9 +137,13 @@ impl Order {
             .await?;
 
         self.nonce = nonce_from_response(&rsp);
-        let body = hyper::body::to_bytes(Problem::from_response(rsp).await?).await?;
+        let body_bytes = Problem::from_response(rsp)
```
djc, line 140: Let's keep this named as `body` at least in this commit to avoid unrelated changes.

src/lib.rs
```diff
@@ -448,16 +455,20 @@ struct Client {
...
+            .uri(directory_url)
+            .body(
+                http_body_util::Empty::new()
+                    .map_err::<_, Error>(|_| unreachable!("Should be Infallible"))
+                    .boxed(),
+            )
+            .expect("Infallible error should not occur");
         let rsp = http.request(req).await?;
-        let body = hyper::body::to_bytes(rsp.into_body()).await?;
+        let body_bytes = rsp.into_body().collect().await?.to_bytes();
```
djc, line 468: No need to rename this.

src/lib.rs
```diff
@@ -467,15 +478,18 @@ impl Client {
         nonce: Option<String>,
         signer: &impl Signer,
         url: &str,
-    ) -> Result<Response<Body>, Error> {
+    ) -> Result<Response<hyper::body::Incoming>, Error> {
```
djc, line 481: You've imported `Incoming` already, no need to qualify it here.

src/lib.rs
```diff
@@ -467,15 +478,18 @@ impl Client {
...
         let body = JoseJson::new(payload, signer.header(Some(&nonce), url), signer)?;
         let request = Request::builder()
             .method(Method::POST)
             .uri(url)
             .header(CONTENT_TYPE, JOSE_JSON)
-            .body(Body::from(serde_json::to_vec(&body)?))
-            .unwrap();
+            .body(
+                http_body_util::Full::from(serde_json::to_vec(&body)?)
+                    .map_err::<_, Error>(|_| unreachable!("Should be Infallible"))
```
djc, line 490: I'd prefer to just stick with `unwrap()` instead of this `map_err()` contraption.

src/lib.rs
```diff
@@ -650,21 +668,35 @@ impl Signer for ExternalAccountKey {
...
         .get(REPLAY_NONCE)
         .and_then(|hv| String::from_utf8(hv.as_ref().to_vec()).ok())
 }
 
 #[cfg(feature = "hyper-rustls")]
-struct DefaultClient(hyper::Client<hyper_rustls::HttpsConnector<HttpConnector>>);
+struct DefaultClient(
+    hyper_util::client::legacy::Client<
+        hyper_rustls::HttpsConnector<HttpConnector>,
+        BoxBody<Bytes, Error>,
```
djc, line 681: Why do we need to use `BoxBody` here? What are we abstracting over?

djc (commented):
Thanks for working on this! I'm afraid this will need some effort to get to a mergeable state.

src/lib.rs
kellpossible: Is there a way to cast `Body`'s associated `Error` type from `Infallible` to `instant_acme::Error` with `Result::unwrap()`? That's what's happening here.

kellpossible: It's due to a change in how hyper works in 1.0

src/lib.rs
kellpossible: Perhaps we could just use `Full` and provide it with no bytes as a substitute for `Empty`, I'm not sure if that breaks any invariants that `Full` requires, it does seem to go against the grain a little from how the types are expected to be used but I guess it does allow us to avoid trait boxing. Another alternative could be to make our own body enum with `Full` and `Empty` variants using enum dispatch technique.

src/lib.rs
kellpossible: But I'll double check and log an issue for it.

Cargo.toml
kellpossible: Sorry a leftover after attempting to upgrade it and running into multiple breaking changes, I'll drop the `.1`!

kellpossible: Curious though why the patch version is left off from this one, but all the other dependencies include it?

src/lib.rs
djc: Using an empty `Full` seems like a good solution here.

Cargo.toml
djc: Typically when you first add a dependency you take the current version, and then when you do a semver-incompatible bump you leave off the patch version (because `.0` patch versions are meaningless).

kellpossible:
@djc  thanks for the review!

Just wondering there's a failed audit check due to license of a subdependency:

https://github.com/instant-labs/instant-acme/actions/runs/9630966408/job/26563174365#step:4:92

This is due to the enabled `webpki-roots` feature on `hyper-rustls` which is being relied upon to allow [an infallible `Default` implementation  for `DefaultClient`](https://github.com/instant-labs/instant-acme/pull/50/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R713)

What would you like to do about this? We could switch to a fallible replacement for `Default` perhaps instead? (Like a `try_new()` method)

src/lib.rs
```diff
@@ -650,21 +668,35 @@ impl Signer for ExternalAccountKey {
...
     fn request(
         &self,
-        req: Request<Body>,
-    ) -> Pin<Box<dyn Future<Output = hyper::Result<Response<Body>>> + Send>> {
+        req: Request<BoxBody<Bytes, Error>>,
+    ) -> Pin<
+        Box<
+            dyn Future<
+                    Output = Result<
+                        Response<hyper::body::Incoming>,
```
djc, line 694: Use unqualified `Incoming` here, too, please.

src/lib.rs
```diff
@@ -690,19 +726,37 @@ pub trait HttpClient: Send + Sync + 'static {
...
     fn request(
         &self,
-        req: Request<Body>,
-    ) -> Pin<Box<dyn Future<Output = hyper::Result<Response<Body>>> + Send>>;
+        req: Request<BoxBody<Bytes, Error>>,
+    ) -> Pin<
+        Box<
+            dyn Future<
+                    Output = Result<
+                        Response<hyper::body::Incoming>,
```
djc, line 734: And here.

djc:
> What would you like to do about this? We could switch to a fallible replacement for `Default` perhaps instead? (Like a `try_new()` method)

Hmm, we should probably start using the rustls-platform-verifier crate instead.

src/lib.rs
kellpossible: https://github.com/instant-labs/instant-acme/issues/53

src/lib.rs
kellpossible: Their `Body` implementations are slightly different, Even when there is no data, [`Full` returns a `Some` with a `Frame` ](https://docs.rs/http-body-util/latest/src/http_body_util/full.rs.html#39-44), and [`Empty` returns a `None`](https://docs.rs/http-body-util/latest/src/http_body_util/empty.rs.html#28-33), I guess it's up to the `HttpClient` implementation whether this would make any difference.

kellpossible: I've swapped to using `Full` anyway, and conveniently it resolved the issue https://github.com/instant-labs/instant-acme/pull/50#discussion_r1650033771

kellpossible:
> It's okay to bump the MSRV to what we need here. If you rebase on top of https://github.com/instant-labs/instant-acme/pull/52 that should fix the audit issue.

@djc I bumped the MSRV to 1.64

src/lib.rs
```diff
@@ -451,10 +457,10 @@ impl Client {
     async fn new(server_url: &str, http: Box<dyn HttpClient>) -> Result<Self, Error> {
         let req = Request::builder()
             .uri(server_url)
-            .body(Body::empty())
-            .unwrap();
+            .body(http_body_util::Full::default())
```
djc, line 460: Nit: no need to qualify `Full`.

src/lib.rs
```diff
@@ -467,15 +473,14 @@ impl Client {
...
+    ) -> Result<Response<Incoming>, Error> {
         let nonce = self.nonce(nonce).await?;
         let body = JoseJson::new(payload, signer.header(Some(&nonce), url), signer)?;
         let request = Request::builder()
             .method(Method::POST)
             .uri(url)
             .header(CONTENT_TYPE, JOSE_JSON)
-            .body(Body::from(serde_json::to_vec(&body)?))
-            .unwrap();
+            .body(http_body_util::Full::from(serde_json::to_vec(&body)?))?;
```
djc, line 483: Same here.

src/lib.rs
```diff
@@ -488,8 +493,8 @@ impl Client {
         let request = Request::builder()
             .method(Method::HEAD)
             .uri(&self.urls.new_nonce)
-            .body(Body::empty())
-            .unwrap();
+            .body(http_body_util::Full::default())
```
djc, line 496: And here.

src/lib.rs
```diff
@@ -650,21 +655,28 @@ impl Signer for ExternalAccountKey {
...
+fn nonce_from_response(rsp: &Response<Incoming>) -> Option<String> {
     rsp.headers()
         .get(REPLAY_NONCE)
         .and_then(|hv| String::from_utf8(hv.as_ref().to_vec()).ok())
 }
 
 #[cfg(feature = "hyper-rustls")]
-struct DefaultClient(hyper::Client<hyper_rustls::HttpsConnector<HttpConnector>>);
+struct DefaultClient(
+    hyper_util::client::legacy::Client<hyper_rustls::HttpsConnector<HttpConnector>, Full<Bytes>>,
```
djc, line 666: Can we import `hyper_util::client::legacy::Client as LegacyClient`?

src/types.rs
```diff
@@ -119,21 +128,23 @@ pub struct Problem {
...
 
-    pub(crate) async fn from_response(rsp: Response<Body>) -> Result<Body, Error> {
+    pub(crate) async fn from_response(
+        rsp: Response<Incoming>,
+    ) -> Result<BoxBody<Bytes, Error>, Error> {
         let status = rsp.status();
         let body = rsp.into_body();
         if status.is_informational() || status.is_success() || status.is_redirection() {
-            return Ok(body);
+            return Ok(body.map_err::<_, Error>(|_| unreachable!()).boxed());
```
djc, line 143: Adjust this like the other ones?

src/types.rs
```diff
@@ -119,21 +128,23 @@ pub struct Problem {
...
         }
 
-        let body = hyper::body::to_bytes(body).await?;
-        Err(serde_json::from_slice::<Problem>(&body)?.into())
+        let body_bytes = body.collect().await?.to_bytes();
```
djc, line 146: Avoid renaming this.

src/types.rs
djc: Let's call this `Http` and rename the one currently named `Http` to `Hyper`.

src/types.rs
```diff
@@ -30,6 +33,12 @@ pub enum Error {
     /// HTTP request failure
     #[error("HTTP request failure: {0}")]
     Http(#[from] hyper::Error),
+    /// HTTP request failure
+    #[error("HTTP request failure: {0}")]
+    HttpHttp(#[from] hyper::http::Error),
+    /// HTTP client request failure
+    #[error("HTTP request failure: {0}")]
+    HttpClient(#[from] hyper_util::client::legacy::Error),
```
djc, line 41: I don't want `hyper_util` to appear in the public API, let's do an `Other<Box<dyn StdError + Send + Sync + 'static>>`?

src/lib.rs
```diff
@@ -690,19 +706,33 @@ pub trait HttpClient: Send + Sync + 'static {
     /// Send the given request and return the response
     fn request(
         &self,
-        req: Request<Body>,
-    ) -> Pin<Box<dyn Future<Output = hyper::Result<Response<Body>>> + Send>>;
+        req: Request<Full<Bytes>>,
+    ) -> Pin<
+        Box<
+            dyn Future<Output = Result<Response<Incoming>, hyper_util::client::legacy::Error>>
```
djc, line 712: I don't think it makes sense to have this very specific error appear in this trait. Maybe just make it `Error`?

djc:
> If the license check with webpki-roots included now passes due to these changes, can we consider this [#50 (comment)](https://github.com/instant-labs/instant-acme/pull/50#issuecomment-2186043352) resolved too, without needing to use rustls-platform-verifier yet? Perhaps can be done in next MR?

No, I don't want to merge a dependency on webpki-roots. Instead you can make the `HttpClient` constructor fallible.

src/lib.rs
kellpossible: I made it `HyperClient` let me know if that's okay!

src/lib.rs
kellpossible: Sure let's do that then, and if we add the Other variant that gives more freedom for the implementer?

kellpossible:
@djc thanks again! I think I may have addressed all the outstanding comments?

src/lib.rs
```diff
@@ -9,11 +9,13 @@ use std::pin::Pin;
...
-#[cfg(feature = "hyper-rustls")]
-use hyper::client::HttpConnector;
+use http_body_util::{BodyExt, Full};
+use hyper::body::{Bytes, Incoming};
 use hyper::header::{CONTENT_TYPE, LOCATION};
-use hyper::{Body, Method, Request, Response, StatusCode};
+use hyper::{Method, Request, Response, StatusCode};
+#[cfg(feature = "hyper-rustls")]
+use hyper_util::client::legacy::connect::HttpConnector;
+use hyper_util::client::legacy::{connect::Connect, Client as HyperClient};
```
djc, line 18: Tiny nit: let's keep these two on separate lines (one line per module like the rest of these).

src/lib.rs
```diff
@@ -451,10 +457,10 @@ impl Client {
     async fn new(server_url: &str, http: Box<dyn HttpClient>) -> Result<Self, Error> {
         let req = Request::builder()
             .uri(server_url)
-            .body(Body::empty())
-            .unwrap();
+            .body(Full::default())
+            .expect("Infallible error should not occur");
```
djc, line 461: Tiny nit: make all error messages lowercase, please.

src/lib.rs
```diff
@@ -488,8 +493,8 @@ impl Client {
         let request = Request::builder()
             .method(Method::HEAD)
             .uri(&self.urls.new_nonce)
-            .body(Body::empty())
-            .unwrap();
+            .body(Full::default())
+            .expect("Should be Infallible");
```
djc, line 497: Lowercase.

src/lib.rs
```diff
@@ -650,38 +655,40 @@ impl Signer for ExternalAccountKey {
...
 
 #[cfg(feature = "hyper-rustls")]
-impl Default for DefaultClient {
-    fn default() -> Self {
-        Self(
-            hyper::Client::builder().build(
+impl DefaultClient {
+    fn try_new() -> Result<Self, Error> {
+        Ok(Self(
+            HyperClient::builder(hyper_util::rt::TokioExecutor::new()).build(
```
djc, line 671: Nit: import `TokioExecutor`, please.

src/types.rs
```diff
@@ -27,9 +29,12 @@ pub enum Error {
     /// Failed to instantiate a private key
     #[error("invalid key bytes: {0}")]
     CryptoKey(#[from] ring::error::KeyRejected),
-    /// HTTP request failure
+    /// Hyper request failure
     #[error("HTTP request failure: {0}")]
-    Http(#[from] hyper::Error),
+    Hyper(#[from] hyper::Error),
+    /// HTTP failure
+    #[error("HTTP request failure: {0}")]
```
djc, line 36: Nit: move `Http` before `Hyper`.

src/types.rs
```diff
@@ -39,6 +44,9 @@ pub enum Error {
     /// Miscellaneous errors
     #[error("missing data: {0}")]
     Str(&'static str),
+    /// Other kind of error
+    #[error(transparent)]
+    Other(Box<dyn std::error::Error + Send + Sync + 'static>),
```
djc, line 49: Nit: let's keep these in alphabetic order.

djc (approved):
This is looking good, a few more small things to address.

Can you also bump the version to 0.6.0?

djc:
Also need to address the CI issues:

* Bump the MSRV in the CI workflow to 1.64
* Add `BSD-3-Clause` to the allowed licenses in `deny.toml`
* Fix the clippy lint

kellpossible:
@djc done!

djc (approved):

src/lib.rs
djc: Yes, that's better!

djc:
Thanks for your patience working through my feedback!

* Published instant-acme v0.6.0 at registry `crates-io`
* [new tag]           0.6.0 -> 0.6.0

gahooa:
@djc and @kellpossible, just wanted to thank you both for getting this done.   We are in the middle of building management features for around 10,000 certificates, and being able to work with great maintainers and developers is so crucial.  Thank you again!

djc:
Always curious to get a bit more context about how/where you're using instant-acme!

gahooa:
@djc I can give a brief overview here, and if you'd like more details, reach out to me at appcove.

We have several clients who operate saas type systems for thousands and thousands of businesses.  These business oftentimes bring their own domains, or pick a subdomain, and some of them generated a multitude of new sites on a daily basis.

There are some great tools out there for automatic TLS issuance, like the caddy webserver.  The issue with such tools is that while they fully automate the process, they provide little central insight or control into how it's going.  Also, we do not like the idea of giving each web-server potentially destructive DNS API access in order to respond to DNS challenges.  This calls for something more centralized.

We are building a central infrastructure management solution with TLS certificate audits/issuance/distribution being one aspect.   It will control when and how certificates are renewed, be able to audit that, report on issues, and so on and so forth. Then on the webserver side, we are working to ensure that the ALPN verification challenges can be successfully answered in real-time by communicating with the central management system.   

This is the area that instant-acme helps us -- to be the workhorse of issuance and renewals across a number of providers, both with DNS and ALPN, and to provide strongly typed rust APIs for accessing both the challenge information and the issued certificates, so they can both be appropriately distributed to the respective web server clusters.

That's it in a nutshell. We appreciate your efforts to make this possible.

PR #54, opened 2024-06-27
Fix #51 handle unknown challenge types

### LINKED ISSUE #51: Future proof the parsing of challenge types
When testing this with pebble https://github.com/letsencrypt/pebble this library fails to parse some of the challenge types that it offers. It would be good to make this enum `#[non_exhaustive]` and perhaps have an option for parsing the list of challenges to either ignore unknown types, or alternatively we could add a new variant something like:

```rust
#[allow(missing_docs)]
#[non_exhaustive] // It's likely more variants may be added in the future.
pub enum ChallengeType {
    #[serde(rename = "http-01")]
    Http01,
    #[serde(rename = "dns-01")]
    Dns01,
    TlsAlpn01,
    Unknown(String),
}
```

kellpossible:
Fix #51 handle unknown challenge types

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -458,7 +458,7 @@ pub enum Identifier {
 }
 
 /// The challenge type
-#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
 #[allow(missing_docs)]
 pub enum ChallengeType {
     #[serde(rename = "http-01")]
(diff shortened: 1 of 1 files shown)


djc:
Ah, should've held off on that semver-breaking release... Oh, well.

Would it be hard to add a small unit test to verify that this does the right thing in the context of JSON as returned by an ACME server?

djc:
I'm incorporating this in #57 (retaining your authoring of the relevant commit). Thanks!

PR #56, opened 2024-07-14
Add aws_lc_rs feature and set it as default

SajjadPourali:
This PR adds the aws_lc_rs feature to the instant-acme library, sets it as the default feature, and updates the dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,9 @@ categories = ["web-programming", "api-bindings"]
 
 [features]
 default = ["hyper-rustls", "aws_lc_rs"]
-ring = ["dep:ring", "hyper-rustls/ring", "rcgen/ring"]
-aws_lc_rs = ["dep:aws-lc-rs", "hyper-rustls/aws-lc-rs", "rcgen/aws_lc_rs"]
+ring = ["dep:ring", "hyper-rustls?/ring", "rcgen/ring"]
+aws_lc_rs = ["dep:aws-lc-rs", "hyper-rustls?/aws-lc-rs", "rcgen/aws_lc_rs"]
+fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
 
 [dependencies]
 base64 = "0.22.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ categories = ["web-programming", "api-bindings"]
 [features]
 default = ["hyper-rustls", "ring"]
 ring = ["dep:ring", "hyper-rustls?/ring", "rcgen/ring"]
-aws_lc_rs = ["dep:aws-lc-rs", "hyper-rustls?/aws-lc-rs", "rcgen/aws_lc_rs"]
-fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
+aws-lc-rs = ["dep:aws-lc-rs", "hyper-rustls?/aws-lc-rs", "rcgen/aws_lc_rs"]
+fips = ["aws-lc-rs", "aws-lc-rs?/fips"]
 
 [dependencies]
 aws-lc-rs = { version = "1.8.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,27 +12,47 @@ keywords = ["letsencrypt", "acme"]
 categories = ["web-programming", "api-bindings"]
 
 [features]
-default = ["hyper-rustls"]
+default = ["hyper-rustls", "aws_lc_rs"]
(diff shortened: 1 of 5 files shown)


SajjadPourali:
To fix the test error, I disabled the test when both `ring` and `aws_lc_rs` features are enabled. The TLS client cannot choose a default provider when both are enabled.

See https://github.com/rustls/rustls/blob/main/rustls/src/crypto/mod.rs#L270-L283

djc:
I'm happy to add a feature for using aws-lc-rs, but I want to stick with ring as the default for now and please avoid any changes unrelated to this (such as TOML reformatting and version bumps to other dependencies).

src/lib.rs
```diff
@@ -13,14 +17,13 @@ use http_body_util::{BodyExt, Full};
 use hyper::body::{Bytes, Incoming};
 use hyper::header::{CONTENT_TYPE, LOCATION};
 use hyper::{Method, Request, Response, StatusCode};
+use hyper_util::client::legacy::{connect::Connect, Client as HyperClient};
```
djc, line 20: I prefer to stick to one line per module (so keeping the separate lines for `Connect` and `Client`, as before).

src/types.rs
```diff
@@ -1,15 +1,18 @@
...
-use ring::signature::{EcdsaKeyPair, KeyPair};
+#[cfg(all(feature = "ring", not(feature = "aws_lc_rs")))]
+pub(crate) use ring as ring_like;
+use ring_like::digest::{digest, Digest, SHA256};
+use ring_like::signature::{EcdsaKeyPair, KeyPair};
 use rustls_pki_types::CertificateDer;
 use serde::de::DeserializeOwned;
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Serialize};
+use std::fmt;
```
djc, line 15: This should remain in a separate block of imports by itself, as before.

Cargo.toml
```diff
@@ -29,7 +33,7 @@ thiserror = "1.0.30"
 [dev-dependencies]
 anyhow = "1.0.66"
 clap = { version = "4.0.29", features = ["derive"] }
-rcgen = "0.12"
+rcgen = { version = "0.12", features = ["pem"], default-features = false }
```
djc, line 36: As elsewhere, please keep `default-features = false` before `features = []`.

djc (commented):
That looks a lot better. There are still a few small unrelated changes that I'd prefer to avoid.

Cargo.toml
```diff
@@ -12,15 +12,19 @@ keywords = ["letsencrypt", "acme"]
 categories = ["web-programming", "api-bindings"]
 
 [features]
-default = ["hyper-rustls"]
+default = ["hyper-rustls", "ring"]
+ring = ["dep:ring", "hyper-rustls?/ring", "rcgen/ring"]
+aws_lc_rs = ["dep:aws-lc-rs", "hyper-rustls?/aws-lc-rs", "rcgen/aws_lc_rs"]
```
djc, line 17: Would prefer to name the feature with hyphens instead of underscores.

djc:
CI is failing because `--all-features` now enables `fips`, which only works on Linux. Probably makes sense to change CI to do the following:

- `check --no-default-features`
- `test --features hyper-rustls,aws-lc-rs`
- `test --features hyper-rustls,ring`

SajjadPourali:
> CI is failing because `--all-features` now enables `fips`, which only works on Linux. Probably makes sense to change CI to do the following:
> 
>     * `check --no-default-features`
> 
>     * `test --features hyper-rustls,aws-lc-rs`
> 
>     * `test --features hyper-rustls,ring`

The suggested commands don’t work because ring is in the default feature.
What about the following?
* `cargo check`
* `cargo test`
* `cargo test --no-default-features --features hyper-rustls,aws_lc_rs`

SajjadPourali:
> CI is failing because `--all-features` now enables `fips`, which only works on Linux. Probably makes sense to change CI to do the following:
> 
>     * `check --no-default-features`
> 
>     * `test --features hyper-rustls,aws-lc-rs`
> 
>     * `test --features hyper-rustls,ring`

The suggested commands don’t work because ring is in the default feature.
What about the following?
* `cargo check`
* `cargo test`
* `cargo test --no-default-features --features hyper-rustls,aws_lc_rs`

djc:
Ah, sorry, that sounds good!

SajjadPourali:
@djc Done. Apparently, on Windows, NASM is a dependency for aws_lc_rs. Any idea?

djc (approved):
This is a good start. Going to merge this and follow up in a separate PR.

PR #57, opened 2024-07-16
Prepare 0.7.0 release

### LINKED ISSUE #51: Future proof the parsing of challenge types

djc:
* Followup from #56 to clean things up a bit
* Cleanup from #50 
* Restore ability to build `HttpClient` implementations without hyper (after #50)
* Incorporate #54 (fixes #51)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -552,24 +552,14 @@ impl Key {
         let rng = crypto::SystemRandom::new();
         let pkcs8 =
             crypto::EcdsaKeyPair::generate_pkcs8(&crypto::ECDSA_P256_SHA256_FIXED_SIGNING, &rng)?;
-        #[cfg(all(feature = "ring", not(feature = "aws-lc-rs")))]
-        let key = crypto::EcdsaKeyPair::from_pkcs8(
-            &crypto::ECDSA_P256_SHA256_FIXED_SIGNING,
-            pkcs8.as_ref(),
-            &rng,
-        )?;
-        #[cfg(feature = "aws-lc-rs")]
-        let key = crypto::EcdsaKeyPair::from_pkcs8(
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -800,9 +800,9 @@ pub trait BytesBody {
 }
 
 mod crypto {
-    #[cfg(all(feature = "aws-lc-rs", not(feature = "ring")))]
+    #[cfg(all(feature = "aws-lc-rs", any(feature = "fips", not(feature = "ring"))))]
     pub(crate) use aws_lc_rs as ring_like;
-    #[cfg(feature = "ring")]
+    #[cfg(all(feature = "ring", not(feature = "fips")))]
```
valkum: I think this is worth some explanations in the README.
As I understand it:
If `ring` and `aws-lc-rs` are enabled: `ring` takes precedence unless `fips` is also enabled, in which case `aws-lc-rs` takes precedence.

valkum (approved):
Only a small recommendation. Looks good otherwise.

src/lib.rs
djc: That's fair, added a commit for this.

valkum (approved):

PR #65, opened 2024-09-16
Retry bad nonce errors with a new nonce

nick96:
I'm putting this up as a possible solution for handling bad nonce errors. The new `BytesBody` implementation feels a bit weird but I couldn't think of a better way that wouldn't have wide reaching interface changes. I think it makes sense to do the retry within `Client` so that everything benefits from it. The main thing this is missing is some backoff, though that may not even be necessary because we're not retrying due to a network or server side error.

Fixes: https://github.com/djc/instant-acme/issues/62

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -492,6 +492,43 @@ impl Client {
     async fn post(
         &self,
         payload: Option<&impl Serialize>,
+        mut nonce: Option<String>,
+        signer: &impl Signer,
+        url: &str,
+    ) -> Result<BytesResponse, Error> {
+        let mut retries = 3;
+        loop {
+            let mut response = self
+                .post_attempt(payload, nonce.clone(), signer, url)
+                .await?;
+            if response.parts.status != StatusCode::BAD_REQUEST {
+                return Ok(response);
+            }
+            let body = response.body.into_bytes().await.map_err(Error::Other)?;
+            let problem = serde_json::from_slice::<Problem>(&body)?;
+            if let Some("urn:ietf:params:acme:error:badNonce") = problem.r#type.as_deref() {
+                retries -= 1;
+                if retries != 0 {
+                    // Retrieve the new nonce. If it isn't there (it
+                    // should be, the spec requires it) then we will
+                    // manually refresh a new one in `post_attempt`
+                    // due to `nonce` being `None` but getting it from
+                    // the response saves us making that request.
+                    nonce = nonce_from_response(&response);
+                    continue;
+                }
+            }
+
+            return Ok(BytesResponse {
+                parts: response.parts,
+                body: Box::new(body),
+            });
+        }
+    }
+
+    async fn post_attempt(
+        &self,
+        payload: Option<&impl Serialize>,
         nonce: Option<String>,
         signer: &impl Signer,
         url: &str,
@@ -786,6 +823,13 @@ where
             Ok(body) => Ok(body.to_bytes()),
             Err(e) => Err(e.into()),
         }
+    }
+}
+
+#[async_trait]
+impl BytesBody for Bytes {
+    async fn into_bytes(&mut self) -> Result<Bytes, Box<dyn StdError + Send + Sync + 'static>> {
+        Ok(self.to_owned())
     }
 }
 


nick96:
Just realised that I don't need to do the clone of `BytesResponse.parts` at all. I'll squash the commits into one much more descriptive one if you approve this!

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
         nonce: Option<String>,
         signer: &impl Signer,
         url: &str,
+    ) -> Result<BytesResponse, Error> {
+        let mut retries = 3;
+
+        let mut nonce = nonce;
```
djc, line 501: I don't think there is a need to rebind `nonce`? We can just define the argument as `mut nonce`?

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
...
+        loop {
+            let mut response = self.inner_post(payload, nonce.clone(), signer, url).await?;
+            if response.parts.status == StatusCode::BAD_REQUEST {
```
djc, line 504: Nit: please invert the condition so we can do the early return here and reduce rightward drift in the rest of the loop body.

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
...
+                let body = response.body.into_bytes().await.map_err(Error::Other)?;
+                let problem: Problem = serde_json::from_slice(&body)?;
```
djc, line 506: Nit: I prefer to avoid type annotation, in favor of turbofishing where necessary.

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
...
+                if problem
+                    .r#type
+                    .map(|t| t == "urn:ietf:params:acme:error:badNonce")
+                    .unwrap_or_default()
```
djc, line 510: Could this just be `if let Some("urn:ietf:params:acme:error:badNonce") = &problem.r#type {` instead?

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
...
+                {
+                    retries -= 1;
+                    if retries != 0 {
+                        nonce = Some(self.nonce(None).await?);
```
djc, line 514: Acquiring a nonce here seems unnecessary -- we can just set it `None`, since `inner_post()` will take care of this for us?

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
...
+                        continue;
+                    }
+                }
+
+                let response = BytesResponse {
+                    parts: response.parts,
+                    body: Box::new(StaticBody { bytes: body }),
+                };
```
djc, line 522: Please inline the `response` value into the `return` statement.

src/lib.rs
```diff
@@ -495,6 +495,43 @@ impl Client {
...
+                return Ok(response);
+            }
+            return Ok(response);
+        }
+    }
+
+    async fn inner_post(
```
djc, line 529: Nit: let's call this `post_attempt()`?

src/lib.rs
```diff
@@ -789,6 +826,17 @@ where
     }
 }
 
+struct StaticBody {
```
djc, line 829: Instead of introducing a wrapper, I think we could just implement `BytesBody` for `Bytes` itself?

djc (commented):
This is looking great, thanks! A whole bunch of minor suggestions follow, please squash everything into one commit which I think I could release shortly.

djc:
This is great, thanks!

src/lib.rs
nick96: I realised that a new nonce is available in the `Replay-Nonce` header so have switched to grabbing that out instead. It fails gracefully, setting `nonce` to `None` if the header isn't there, but the header is required by the spec so should definitely be there. This way we avoid an HTTP request all together.

nick96:
Sorry, I didn't realise you were actively reviewing and force pushed a change to use the `Replay-Nonce` header.

djc:
No problem!

src/lib.rs
```diff
@@ -490,6 +490,47 @@ impl Client {
...
+                    // Retrieve the new nonce from the header. If it
+                    // isn't there (it should be, the spec requires
+                    // it) then we will manually refresh a new one in
+                    // `post_attempt` due to `nonce` being `None` but
+                    // using the header saves us a request.
+                    nonce = response
+                        .parts
+                        .headers
+                        .get("Replay-Nonce")
+                        .and_then(|v| v.to_str().map(|s| s.to_string()).ok());
```
djc, line 521: I think this is `nonce_from_response()`? Can we use that here?

nick96: Yeah it is. Done!

djc:
Thanks!

PR #68, opened 2024-10-23
Add `pub fn id()` to Account

kvinwang:
This PR adds `pub fn id()` to Account. The account id is used when we need to add the account to [CAA](https://letsencrypt.org/docs/caa/) records.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -408,6 +408,11 @@ impl Account {
         let _ = Problem::from_response(rsp).await?;
         Ok(())
     }
+
+    /// Get the account ID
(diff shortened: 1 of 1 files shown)


djc (approved):
Would you like a release for this? If so, please submit a PR with a semver-compatible version bump.

kvinwang:
> Would you like a release for this? If so, please submit a PR with a semver-compatible version bump.

I would appreciate having a release. However, I've noticed the version has already been incremented to 0.7.2 which hasn't been published yet. It would be preferable if you manage the version numbering.

PR #74, opened 2025-01-26
Start implementing Pebble-based integration testing

cpu:
:wave: Here's the start of some HTTP-01 integration testing w/ [Pebble](https://github.com/letsencrypt/pebble) as the ACME CA, and the [Pebble challenge test server](https://github.com/letsencrypt/pebble/blob/main/cmd/pebble-challtestsrv/README.md) serving mock DNS and the HTTP-01 response we populate from what instant-acme provides. It should be fairly straight forward to extend to testing other challenges, and different protocol features (e.g. errors with subproblems, ARI, cert profiles) but I wanted to get the minimum viable implementation up for review before going too far. WDYT?

So far a test issuance is performed by:

* Creating an ACME account, and order for "example.com"
* Configuring an HTTP-01 challenge response using the `pebble-challtestsrv`
* Finalizing the order and fetching the certificate chain
* Using the randomly generated Pebble issuer CA root to verify the issued certificate chain w/ rustls/webpki


tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
...
+rcgen = { version = "0.13", features = ["pem"] }
+---
+
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::env;
+
+use rcgen::{DistinguishedName, KeyPair, DnType, IsCa, BasicConstraints};
+
+fn main() {
```
djc, line 15: Suggest changing this to `fn main() -> anyhow::Result<()>` and using `?` instead of `.unwrap()` all over the place.

tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env -S cargo -Z script
```
djc, line 1: I think I'd prefer to keep this in `examples` instead of using the nightly stuff, maybe with a docstring to make it clear that it's not actually an example?

tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use rcgen::{DistinguishedName, KeyPair, DnType, IsCa, BasicConstraints};
+
+fn main() {
+    let ca_key = KeyPair::generate().unwrap();
+    let mut ca_params = rcgen::CertificateParams::default();
+    let mut distinguished_name = DistinguishedName::new();
+    distinguished_name.push(
+        DnType::CommonName,
+        "Pebble CA".to_string(),
```
djc, line 21: Nit: `.to_string()` -> `.to_owned()` to avoid invoking all the `Display` machinery.

tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
...
+    );
+    ca_params.distinguished_name = distinguished_name;
```
djc, line 23: Nit: initialize `ca_params` after setting up `distinguished_name`, instead of the interspersing that happens here?

tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
...
+    ca_params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+
+    let ca_cert = ca_params.self_signed(&ca_key).unwrap();
+    let ca_cert_pem = ca_cert.pem();
+
+    let ca_cert_file_path = env::current_dir().unwrap().join("tests/testdata/ca.pem");
+    let mut ca_cert_file = OpenOptions::new().write(true).create(true).open(ca_cert_file_path).unwrap();
+    ca_cert_file.write_all(ca_cert_pem.as_bytes()).unwrap();
```
djc, line 31: Maybe leverage https://doc.rust-lang.org/std/fs/fn.write.html instead of doing this whole song and dance (same for the other files)?

tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let ee_key = KeyPair::generate().unwrap();
+    let mut ee_params = rcgen::CertificateParams::new(["localhost".to_string(), "127.0.0.1".to_string()]).unwrap();
```
djc, line 34: (Not feedback on this PR, but we should really use something other than `String` for this argument in rcgen... maybe `pki_types::ServerName`?)

tests/testdata/generate.rs
```diff
@@ -0,0 +1,47 @@
...
+    let ee_cert_pem = ee_cert.pem();
+
+    let ee_cert_file_path = env::current_dir().unwrap().join("tests/testdata/server.pem");
+    let mut ee_cert_file = OpenOptions::new().write(true).create(true).open(ee_cert_file_path).unwrap();
+    ee_cert_file.write_all(ee_cert_pem.as_bytes()).unwrap();
+
+    let ee_key_file_path = env::current_dir().unwrap().join("tests/testdata/server.key");
+    let mut ee_key_file = OpenOptions::new().write(true).create(true).open(ee_key_file_path).unwrap();
+    ee_key_file.write_all(ee_key.serialize_pem().as_bytes()).unwrap();
+}
```
djc, line 47: Nit: please add the trailing newline.

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+        format!("https://{}/dir", &self.config.pebble.listen_address)
+    }
+}
+
+#[derive(Clone, Serialize)]
+struct Config {
+    pebble: PebbleConfig,
+}
+
+#[derive(Clone, Serialize)]
```
djc, line 350: Can we use `serde(rename_all = "camelCase")` for this?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+struct PebbleConfig {
```
djc, line 351: Given that this is `Serialize` only, can we get away with using `&str` instead of `String` for all or most of these things (maybe using `Cow<'a, str>` for the remainder)? Would be nice if that means we can directly const-initialize a `PebbleConfig`, without going through `LazyLock`.

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+    for i in 0..10 {
+        if TcpStream::connect(addr).is_ok() {
+            return;
+        }
+        thread::sleep(Duration::from_millis(i * 100));
+    }
+    panic!("failed to connect to {:?} after 10 tries", addr);
+}
+
+static DEFAULT_CONFIG: LazyLock<Config> = LazyLock::new(|| Config {
```
djc, line 413: The use of `LazyLock` means this can't work with the current MSRV, I think? That might be okay I guess to assume we'll only run this on stable?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+struct Config {
+    pebble: PebbleConfig,
+}
+
+#[derive(Clone, Serialize)]
+struct PebbleConfig {
+    #[serde(rename = "listenAddress")]
+    listen_address: String,
+    #[serde(rename = "managementListenAddress")]
+    management_listen_address: String,
```
djc, line 355: Could we use `SocketAddr` for these?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+    external_account_binding_required: bool,
+    #[serde(rename = "domainBlocklist")]
+    domain_blocklist: Vec<String>,
+    retry_after: RetryConfig,
+    profiles: HashMap<String, Profile>,
+}
+
+#[derive(Clone, Serialize)]
+struct RetryConfig {
+    authz: u32,
```
djc, line 375: Maybe add a docstring to specify the units, or use `Duration` and a `serde(serialize_with = "")`?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+impl Drop for Subprocess {
+    fn drop(&mut self) {
+        if let Some(mut child) = self.0.take() {
+            child.kill().expect("failed to kill subprocess");
+            child.wait().expect("failed to wait for killed subprocess");
+        }
+    }
+}
```
djc, line 401: Why do we need these shenanigans? Add some comments?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+
+fn wait_for_server(addr: &str) {
```
djc, line 403: Why not make this `async`?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+        Ok(())
+    }
+
+    // Return a RootCertStore containing the issuer root for the Pebble CA.
+    //
+    // This is distinct from the management root CA, and is randomly generated each
+    // time that Pebble starts up. This is the issuer that signs the randomly generated
+    // intermediate certificate returned as part of ACME issued certificate chains.
+    async fn issuer_roots(&self) -> Result<RootCertStore, Box<dyn StdError>> {
+        let client = self.http_client();
```
djc, line 308: Nit: `fn http_client()` should come after this method.

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+        token: &str,
+        key_auth: &str,
+    ) -> Result<(), Box<dyn StdError>> {
+        let client = self.http_client();
+
+        let body = json!({
+            "token": token,
+            "content": key_auth,
+        })
+        .to_string();
```
djc, line 289: How do you feel about defining an actual `Serialize`-implementing type for this instead of using `json!()`?

Maybe serialize directly into a `Bytes` instead of converting via `String`?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+        sleep(delay).await;
+        let state = order.refresh().await.unwrap();
+        if let OrderStatus::Ready | OrderStatus::Invalid = state.status {
+            break;
+        }
+
+        delay *= 2;
+        tries += 1;
+        match tries < 10 {
+            true => eprintln!(
```
djc, line 126: Maybe good to set this up with a simple `tracing_subscriber` like

```rust
    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
        .init();
```

and use `tracing::info!()` calls instead of `eprintln!()` here?

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+        })
+        .await?;
+
+    // Issue a certificate w/ HTTP-01 challenge and verify it.
+    test_http1(&pebble, order).await?;
+
+    Ok(())
+}
+
+async fn test_http1(env: &PebbleEnvironment, mut order: Order) -> Result<(), Box<dyn StdError>> {
```
djc, line 67: Maybe this should be a `PebbleEnvironment` method?

.github/workflows/pebble.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Download Pebble
+        run: |
+          curl -L "https://github.com/letsencrypt/pebble/releases/latest/download/pebble-linux-amd64.tar.gz" | tar xz --strip-components=3
+          chmod +x pebble
+      - name: Download Pebble Challenge Test Server
+        run: |
+          curl -L "https://github.com/letsencrypt/pebble/releases/latest/download/pebble-challtestsrv-linux-amd64.tar.gz" | tar xz --strip-components=3
+          chmod +x pebble-challtestsrv
+      - name: Run integration test
+        run: cargo test --test integration
```
djc, line 26: Nit: trailing newline, please.

.github/workflows/pebble.yml
```diff
@@ -2,7 +2,7 @@ name: Pebble
 
 on:
   push:
-    branches: ['main']
+    branches: ['main', '*_ci']
```
djc, line 5: Very nitty, but can we make it `-` instead of `_`?

djc (commented):
Cool stuff! Here's a first round of feedback.

tests/testdata/generate.rs
cpu: > I think I'd prefer to keep this in examples instead of using the nightly stuff, maybe with a docstring to make it clear that it's not actually an example?

Sure, WFM.

> I imagine that might also help avoid these certs become stale over time and needing to be regenerated every x months?

The script is using the rcgen default validity period which is insanely large, so at least that won't be a problem. If you don't value the reproducibility we could also just vendor the cert and key without a generator. There's nothing "special" about the certs except the localhost/127.0.0.1 SAN on the EE.

tests/integration.rs
cpu: >  That might be okay I guess to assume we'll only run this on stable?

That was my thinking (and what the CI addition does), but I think it will fall away if I switch to the static lifetimes as suggested in the other comment. Will circle back depending on how that goes.

tests/integration.rs
cpu: Can you be more specific? I'm not sure what part you think is shenanigans :laughing: The kill and wait? the option take? doing it in drop?

djc: The whole wrapping of the child process and the `Drop` impl.

tests/testdata/generate.rs
cpu: Done.

tests/testdata/generate.rs
cpu: > I think I'd prefer to keep this in examples instead of using the nightly stuff, maybe with a docstring to make it clear that it's not actually an example?

Done.

> Doing it inline in the test also makes a lot of sense

I ended up leaving this as an example generator + vendored certs/key because it let us keep the `static` pebble config. Unfortunately there's no way to provide the management cert/key by command-line flags and so making them dynamically complicates the default config where it needs file paths.

It's a special kind of "fun" to suffer your own code as a user years after writing it. Turns out future me is not a fan of some things past me did :zany_face:

tests/testdata/generate.rs
cpu: Done

tests/testdata/generate.rs
cpu: Done.

tests/testdata/generate.rs
cpu: Much nicer, thanks for the prod.

tests/testdata/generate.rs
cpu: Done. I was trying rust-rover for this and hadn't ported over my usual settings from clion :man_facepalming:

tests/integration.rs
cpu: Thanks! much better.

tests/integration.rs
cpu: Good idea. The only complication was the `HashMap` field for the profiles. The benefit of making this `const` friendly seemed to merit some special contortions so I switched this to use:

```rust
#[serde(serialize_with = "serialize_profiles")]
profiles: &'static [(&'static str, Profile)],
```

and then implemented a small custom serializer for the field.

tests/integration.rs
cpu: This fell away.

tests/integration.rs
cpu: I don't think we can `const` construct a `SocketAddr` so I left this as a `&'static str`. It also feels like it's of less value in this case since we're mostly using this as a string to construct a URL, and not binding the address ourselves outside of the loop that polls for the port to be ready.

tests/integration.rs
cpu: Added docstrings.

tests/integration.rs
```diff
@@ -0,0 +1,433 @@
...
+struct RetryConfig {
+    authz: u32,
+    order: u32,
+}
+
+#[derive(Clone, Serialize)]
+struct Profile {
+    description: String,
+    #[serde(rename = "validityPeriod")]
+    validity_period: u32,
```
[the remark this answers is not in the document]
cpu: Added docstring and renamed the field to `validity_period_seconds`.

tests/integration.rs
cpu: I added some comments. If we spawn the `Child` processes but don't explicitly `kill()` and `wait()` they'll be left running after the test ends. We could do this work manually at the end of `main` but it feels more Rust-y to have this all handled consistently for both sub-processes through a wrapper type and a `Drop` impl. The inspiration comes from [rustls-openssl-compat](https://github.com/cpu/rustls-openssl-compat/blob/5247252014aa62079708b7a9c9095a18d6e3f659/tests/runner.rs#L730-L765) where it does a bit more work w.r.t capturing output.

tests/integration.rs
cpu: Done.

tests/integration.rs
cpu: Done.

tests/integration.rs
cpu: Done x2.

tests/integration.rs
cpu: I moved this and a handful of other free-standing functions to be methods of the environment.

.github/workflows/pebble.yml
cpu: Fixed. Another IDE settings oops.

.github/workflows/pebble.yml
cpu: Changed.

Cargo.toml
```diff
@@ -41,7 +41,7 @@ clap = { version = "4.0.29", features = ["derive"] }
 rcgen = { version = "0.13", default-features = false, features = ["pem"] }
 tokio = { version = "1.22.0", features = ["macros", "rt", "rt-multi-thread", "time"] }
 tracing = "0.1.37"
-tracing-subscriber = "0.3.16"
+tracing-subscriber =  { version = "0.3.16", features = ["env-filter"] }
```
djc, line 46: Nit: would prefer to squash this change into the commit that actually needs it.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
+use std::error::Error as StdError;
+use std::io::Read;
+use std::process::{Child, Command};
+use std::time::Duration;
+use std::{env, fs, io};
```
djc, line 5: Nit: move the `io` import to `std::io::{self, Read}`.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+    /// Create a test environment for the given configuration.
+    ///
+    /// Set the PEBBLE and CHALLTESTSRV to pebble and pebble-challtestsrv binaries
+    /// respectively. If unset "./pebble" and "./pebble-challtestsrv" are used.
+    ///
+    /// Returns only once the Pebble CA server interface is responding.
+    async fn new(config: Config) -> io::Result<Self> {
+        let config_file = NamedTempFile::new()?;
+        let config_json = serde_json::to_string_pretty(&config)?;
+        trace!("pebble config:\n{config_json}\n");
```
djc, line 92: Suggest using named values, like this:

```rust
trace!(config = config_json, "using static config");
```

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+        fs::write(&config_file, config_json)?;
+
+        let pebble_path = env::var("PEBBLE").ok().unwrap_or("./pebble".to_owned());
```
djc, line 95: Nit: use `unwrap_or_else()` here?

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+        let challtestsrv_path = env::var("CHALLTESTSRV")
+            .ok()
+            .unwrap_or("./pebble-challtestsrv".to_owned());
```
djc, line 98: And here.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+                .iter()
+                .find(|c| c.r#type == ChallengeType::Http01)
+                .ok_or_else(|| "no http01 challenge found")?;
+
+            let Identifier::Dns(identifier) = &authz.identifier;
+
+            // Provision the HTTP-01 challenge response with the Pebble challenge test server.
+            let key_auth = order.key_authorization(challenge);
+            debug!(
+                "provisioning HTTP-01 response: token={token} key_auth={key_auth}",
```
djc, line 163: Nit: use structured logging instead of faux structure in the message.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+    async fn add_http01_response<'token, 'key_auth>(
+        &self,
+        token: &'token str,
+        key_auth: &'key_auth str,
+    ) -> Result<(), Box<dyn StdError>> {
+        #[derive(Serialize)]
+        struct AddHttp01Request<'token, 'key_auth> {
+            token: &'token str,
+            #[serde(rename = "content")]
+            key_auth: &'key_auth str,
```
djc, line 200: Nit: let's just call the field `content`, please.

(But maybe consider inlining the `body` binding, which feels fairly trivial?)

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+        Ok(())
+    }
+
+    /// Poll the given order until it is ready, waiting longer between each attempt up to
+    /// a maximum.
+    ///
+    /// Returns an error when the maximum number of tries has been reached.
+    async fn poll_until_ready(order: &mut Order) -> Result<(), Box<dyn StdError>> {
+        let url = order.url().to_owned();
+        info!("waiting for order to be ready: {url}");
```
djc, line 221: Nit: structured logging.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+            let state = order.refresh().await.unwrap();
+            if let OrderStatus::Ready | OrderStatus::Invalid = state.status {
+                break;
+            }
+
+            delay *= 2;
+            tries += 1;
+            match tries < 10 {
+                true => info!(
+                    "after {tries} tries order {url} is {status:#?} not ready, waiting {delay:?}",
```
djc, line 235: Nit: structured logging.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+                    status = state.status
+                ),
+                false => {
+                    error!("after {tries} tries - order {url} is not ready: {state:#?}");
```
djc, line 239: Nit: structured logging.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+    /// Issue a certificate for the given order, and identifiers.
+    ///
+    /// The issued certificate chain is verified with the provider roots.
+    async fn issue_certificate(
+        &self,
+        order: &mut Order,
+        identifiers: Vec<String>,
+    ) -> Result<(), Box<dyn StdError>> {
+        info!(
+            "issuing certificate with subject identifiers {identifiers} for order {order_url}",
```
djc, line 261: Nit: structured logging.

djc (commented):
Looking good! I like the logging, but think you should use the structured logging capability tracing provides instead of stuffing everything into the `message` field.

Cargo.toml
cpu: Done

tests/integration.rs
cpu: Done

tests/integration.rs
cpu: Done throughout.

tests/integration.rs
cpu: Switched from `ok().unwrap_or()` to just `unwrap_or()`, good catch. Using `unwrap_or_else()` introduces an unnecessary closure.

tests/integration.rs
cpu: Done.

tests/integration.rs
cpu: Done x2.

tests/integration.rs
```diff
@@ -0,0 +1,495 @@
...
+    /// Sleeps a longer duration after each attempt and panics after 10 failed attempts.
+    async fn wait_for_server(addr: &str) {
+        for i in 0..10 {
+            if TcpStream::connect(addr).await.is_ok() {
+                return;
+            }
+            tokio::time::sleep(Duration::from_millis(i * 100)).await
+        }
+        panic!("failed to connect to {addr:?} after 10 tries");
+    }
```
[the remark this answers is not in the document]
cpu: Done.

tests/integration.rs
djc: So I did actually mean `unwrap_or_else()`, so the allocation from `"./pebble".to_owned()` is only done when necessary. See https://rust-lang.github.io/rust-clippy/master/index.html#/unwrap_or_else.

tests/testdata/generate.rs
djc: Hmm, maybe I got confused. Looks okay to me!

djc (approved):

djc:
Great, thanks!

tests/integration.rs
cpu: Ah, fair enough. Switched.

djc:
Going to do a followup, will tag you for review. 👍

PR #75, opened 2025-01-28
Integration test tweaks

djc:
@cpu happy to get your feedback!


tests/pebble.rs
```diff
@@ -63,8 +63,28 @@ async fn http_01() -> Result<(), Box<dyn StdError>> {
     .await?;
     info!(account_id = account.id(), "created ACME account");
 
-    // Issue a certificate w/ HTTP-01 challenge and verify it.
-    pebble.test_http1(&mut account).await?;
+    // Issue a certificate w/ HTTP-01 challenge.
+    let (identifiers, cert_chain) = pebble.test_http1(&mut account).await?;
+
+    // Split off and parse the EE cert, save the intermediates that follow.
```
cpu, line 69: I think putting this in the test directly is going to lead to duplication for the other challenge types which will want to verify the issued certificate the same way. Maybe lift to a helper fn?

tests/pebble.rs
```diff
@@ -33,8 +33,12 @@ use tracing_subscriber::layer::SubscriberExt;
...
 use tracing_subscriber::{fmt, EnvFilter};
 
-#[tokio::main]
-async fn main() -> Result<(), Box<dyn StdError>> {
+/// Ignored by default because it requires `pebble` and `pebble-challtestsrv` binaries.
+///
+/// See documentation for [`PebbleEnvironment`].
+#[tokio::test]
+#[ignore]
+async fn http_01() -> Result<(), Box<dyn StdError>> {
```
cpu, line 41: I think this is going to be problematic looking forward to adding more tests. My intent was to have `main` spin up one Pebble environment and then serially invoke each test.

I'm not a fan of this approach because if tests are run in parallel there will be port conflicts spinning up multiple pebble environments with the same config. Using a different config per test might be workable, but it ratches up the complexity for little gain IMO.

I was initially thinking of having separate `#[test]`'s but a single global pebble environment but without using something like Nextest I found the Rust unit test runner too primitive to be able to nicely handle setting up/tearing down a global instance reliably.

Sorry, should have put more of this context in the original desc.

cpu (commented):
90% looks great, thank you!

tests/pebble.rs
djc: So IMO the previous setup has a big problem, which is that by default, `cargo c`/`cargo t`/`cargo clippy` don't "see" the code. This makes the IDE experience pretty janky and also resulted in clippy issues slipping through CI. So, I'm open to doing it some other way, but I feel strongly about keeping the code in a place where Cargo can see it by default.

Maybe a solution might be to stick the `PebbleEnvironment` in a `LazyLock` and allow tests to work from there?

tests/pebble.rs
djc: I did consider that. I think it would make sense to lift to a helper `fn` as soon as we actually have more challenge types, but I felt about about keeping the validation in `issue_certificate()` and it seems okay to keep it here until we add multiple challenge types?

tests/pebble.rs
djc: > What about removing `test = false` and then making `main()` skip the test with a message if it doesn't find the required binaries? based on some quick local testing removing that `Cargo.toml` bit seems to do the trick w.r.t clippy.

I did also consider that, but I'm worried that it'll be easy to miss when the tests just don't run, since there is no kind of "skipped" state in the Cargo test framework.

There's a suggestion [here](https://users.rust-lang.org/t/shared-dropable-state-between-tests/70255) that we could use an atomic counter to drop the processes when the counter goes down to zero, maybe that is a feasible solution?

cpu: > There's a suggestion [here](https://users.rust-lang.org/t/shared-dropable-state-between-tests/70255) that we could use an atomic counter to drop the processes when the counter goes down to zero, maybe that is a feasible solution?

That's a decent idea. It sounds workable to me. Do you want to run with it or should I?

djc: I'm running low on time for today. Suggest we merge this and you can pick it up, or I might have a look tomorrow. Got a very sick kid so kinda constrained.

cpu: Sounds good, best of luck to you & kiddo!

cpu (approved):

PR #77, opened 2025-02-11
pebble: improve test environment, test DNS-01 and TLS-ALPN-01 challenge types

cpu:
This replaces https://github.com/djc/instant-acme/pull/76 as a continuation of https://github.com/djc/instant-acme/pull/75.

In #76 I went down the road of trying to manage a single global test environment shared by each individual test, while still ensuring it was properly torn down at the very end. In practice it was pretty gnarly and I'm not happy with where I landed.

Instead of that approach this branch reworks the test methodology so that each unit test can manage its own unique test `Environment`. This should be result in better test isolation and the only required complexity is assigning unique port numbers to the server processes, their various management interfaces, and the challenge types. At the end of each individual test the environment is dropped and the sub-proceses cleaned up.

With this new test logic in place it's fairly trivial to extend test coverage to include DNS-01 and TLS-ALPN-01 challenge types.


tests/pebble.rs
```diff
@@ -334,13 +265,169 @@ impl PebbleEnvironment {
...
+        format!("http://[::1]:{}", self.config.challtestsrv_port)
     }
 
     fn directory_url(&self) -> String {
-        format!("https://{}/dir", &self.config.listen_address)
+        format!("https://{}/dir", &self.config.pebble_config.listen_address)
+    }
+}
+
+/// Configuration for running a `PebbleEnvironment` with unique per-environment ports.
```
djc, line 458: This commit has a lot going on, do you think it could be usefully split into smaller parts?

djc, line 458: This commit has a lot going on, do you think it could usefully be split into smaller parts?

tests/pebble.rs
```diff
@@ -431,6 +410,35 @@ impl PebbleEnvironmentConfig {
     }
 }
 
+fn verify_cert(
```
djc, line 609: Should this be a method on the `pebble`, and internalize the `issuer_roots().await?` call?

tests/pebble.rs
```diff
@@ -185,6 +150,82 @@ impl PebbleEnvironment {
...
     }
 
+    /// Create and finalize an ACME order for the given `Account` and `Identifier`s using
+    /// the specified `ChallengeType`.
+    ///
+    /// The `provision_fn` will be invoked to set up the appropriate challenge response
+    /// based on the `Challenge`/`KeyAuthorization`.
+    ///
+    /// Returns the issued certificate chain unless an error occurs.
+    async fn complete_order<'a>(
```
djc, line 315: I feel like we should probably bite the bullet and figure out how to put minimally opinionated takes on this stuff in the library itself -- not much point in every downstream copy/pasting...

tests/pebble.rs
```diff
@@ -429,7 +469,7 @@ fn verify_cert(
         UnixTime::now(),
         crypto_provider.signature_verification_algorithms.all,
     )
-        .unwrap();
+    .unwrap();
```
djc, line 627: Nit: move this into the original commit?

tests/pebble.rs
```diff
@@ -76,66 +81,26 @@ impl PebbleEnvironment {
...
+                let token = challenge.token.clone();
+                Box::pin(async move {
+                    debug!(
+                        token,
+                        key_auth = key_auth.as_str(),
+                        "provisioning HTTP-01 response",
+                    );
+                    self.add_http01_response(&token, key_auth.as_str()).await
+                })
+            },
```
djc, line 148: Ugh, pretty ugly callback stuff... Maybe an `AuthorizationMethod` trait just for the test?

tests/pebble.rs
```diff
@@ -185,6 +150,82 @@ impl PebbleEnvironment {
...
+        &'a self,
+        account: &mut Account,
+        identifiers: &[String],
+        chal_type: ChallengeType,
+        provision_fn: impl Fn(
+            String,
+            &Challenge,
+            KeyAuthorization,
+        ) -> Pin<
+            Box<dyn Future<Output = Result<(), Box<dyn StdError + 'static>>> + Send + 'a>,
```
djc, line 325: What's the point of boxing the return future?

djc:
Good stuff! I'd really like to find a better way of abstracting over authorization methods...

tests/pebble.rs

tests/pebble.rs

tests/pebble.rs
cpu: I agree, but unless you had a stronger preference I think I'd prefer to land this test code and then take that on as a separate follow-up. WDYT?

tests/pebble.rs

tests/pebble.rs
cpu: That's much cleaner, thanks! It requires declaring some extra structs but I think those make for a good home for some of the misc `add_xxx_response()` fns that I was tacking on to the pebble env before. :+1:

tests/pebble.rs
cpu: This fell away.

tests/pebble.rs
```diff
@@ -296,6 +283,62 @@ impl PebbleEnvironment {
...
+    ) -> Result<(), Box<dyn StdError>>;
+}
+
+/// Provisions an HTTP-01 challenge response.
+///
+/// The Pebble challenge test server at the `management_url` will be configured to respond to
+/// HTTP-01 challenge requests for a provided token by returning the expected key auth.
+struct HttpChallengeProvisioner {
+    management_url: String,
+    client: HyperClient<hyper_rustls::HttpsConnector<HttpConnector>, Full<Bytes>>,
```
djc, line 379: All `ChallengeProvisioner` impls will need a `management_url` and a HTTP `client`, right? Feels like this state should not live on this type... Can we reuse the `PebbleEnvironment` to pass around all of the state here? Or wrap it in a `ChallengeContext` or something?

tests/pebble.rs
```diff
@@ -403,29 +403,55 @@ async fn wait_for_server(addr: &str) {
...
-fn serialize_profiles<S: Serializer>(
-    profiles: &'static [(&'static str, Profile)],
-    serializer: S,
-) -> Result<S::Ok, S::Error> {
-    let mut map = serializer.serialize_map(Some(profiles.len()))?;
-    for (k, v) in profiles {
-        map.serialize_entry(k, v)?;
+impl Default for Config {
+    fn default() -> Self {
+        let profiles = [
```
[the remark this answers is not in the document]

tests/pebble.rs
```diff
@@ -400,6 +412,23 @@ async fn wait_for_server(addr: &str) {
     panic!("failed to connect to {addr:?} after 10 tries");
 }
 
+/// Configuration for running a `PebbleEnvironment` with unique per-environment ports.
+struct PebbleEnvironmentConfig {
```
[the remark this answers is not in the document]
cpu: Done x2. I also added an earlier rename commit to make `PebbleEnvironment` -> `Environment`. I think that works nicer with `EnvironmentConfig` and better encapsulates that it's not just `pebble`.

tests/pebble.rs
```diff
@@ -131,19 +133,29 @@ impl PebbleEnvironment {
...
+                .arg("-management")
+                .arg(format!("[::1]:{}", config.challtestsrv_port))
+                .arg("-dns01")
+                .arg(format!("[::1]:{}", config.dns_port))
+                .arg("-http01")
+                .arg(format!("[::1]:{}", config.pebble_config.http_port))
+                .arg("-tlsalpn01")
+                .arg(format!("[::1]:{}", config.pebble_config.tls_port))
+                .arg("-https01")
+                .arg("") // Disable HTTP-01 over https:// redirect challenges.
```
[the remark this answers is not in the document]
cpu: I left this as-is because I think it makes the commit-by-commit state cleaner. If we only added `-http01` flags here then it gives the appearance all of the refactoring done before was enough to run multiple environments, but it won't be true until each challenge type is added in subsequent commits. In the interim there would be conflicts from the default ports used for the yet-to-be-configured challenge types.

WDYT?

tests/pebble.rs
```diff
@@ -483,6 +484,11 @@ impl Default for PebbleConfig {
     }
 }
 
+fn choose_port() -> u16 {
```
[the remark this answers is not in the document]
cpu: Done. I agree the wrapper wasn't doing much so I dropped it and inlined each increment.

tests/pebble.rs
```diff
@@ -77,15 +80,20 @@ impl PebbleEnvironment {
     async fn test_http1(
         &self,
         account: &mut Account,
-    ) -> Result<(Vec<String>, Vec<CertificateDer<'static>>), Box<dyn StdError>> {
+        identifiers: &[String],
```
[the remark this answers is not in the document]

tests/pebble.rs
cpu: I also added a commit at the end to build the provisioner from the challenge type. I tacked that on separately in case you don't like the approach.

tests/pebble.rs
```diff
@@ -403,29 +403,55 @@ async fn wait_for_server(addr: &str) {
...
     tls_port: u16,
-    ocsp_responder_url: &'static str,
+    ocsp_responder_url: String,
     external_account_binding_required: bool,
-    domain_blocklist: &'static [&'static str],
+    domain_blocklist: Vec<String>,
     retry_after: RetryConfig,
-    #[serde(serialize_with = "serialize_profiles")]
-    profiles: &'static [(&'static str, Profile)],
+    profiles: HashMap<String, Profile>,
```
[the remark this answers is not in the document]
djc: > I don't think `Path`s take a generic lifetime parameter, and just sticking a `Path` in there won't work since the size isn't known at compile time. I think it'd have to be a `PathBuf` (but you want to avoid owned types), or a `&'static Path` (which seems awkward to construct). I left these `&'static str` but could adjust if you have a pref here.

`Path` has `fn new<S: AsRef<OsStr> + ?Sized>(s: &S) -> &Path` which I think means that just `Path::new("foo")` should work to construct a `&'static Path`?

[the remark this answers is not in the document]
djc: > > For everything where it's possible, I think I'd like to just stick to &'static str?
> 
> Given the overall context I was leaning towards consistency here over micro-optimizations 😅

I agree that this might not be entirely rational from a performance perspective. I might however cling to the idea that the type of `&'static str` vs `String` also communicates some semantics in terms of being constructed at runtime vs being compile-time constants which can be useful when reading code.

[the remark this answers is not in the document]
djc: > Profile keys being `&'static str` makes collecting into a `HashMap` require extra contortions that felt like they counteracted the benefit, so I left that as-is. Do you feel strongly about that one?

One more, what contortions? Seems surprising.

djc (approved):
Probably going to do a round of tweaks after this lands, but I don't know exactly where I am going yet. In any case, the improved test coverage here is great!

tests/pebble.rs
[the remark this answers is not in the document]
cpu: > I think I missed that. I'll go back and give using &Path a better shot 👍

It worked great, thanks.

> One more, what contortions? Seems surprising.

I'm not sure what I was stumbling on before. I tried again and had no issues.

Thanks for pushing back on those, I think I was giving up too easily!

tests/pebble.rs
```diff
@@ -420,6 +338,87 @@ struct EnvironmentConfig {
     challtestsrv_port: u16,
 }
 
+impl EnvironmentConfig {
+    /// Consume the `PebbleEnvironmentConfig` and start a Pebble and challenge test servers.
```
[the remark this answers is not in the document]
cpu: Fixed. Thanks!

tests/pebble.rs
```diff
@@ -77,15 +78,20 @@ impl Environment {
...
         account: &mut Account,
-    ) -> Result<(Vec<String>, Vec<CertificateDer<'static>>), Box<dyn StdError>> {
+        identifiers: &[&'static str],
+    ) -> Result<Vec<CertificateDer<'static>>, Box<dyn StdError>> {
         info!("testing HTTP-01 challenge");
 
+        let identifiers = identifiers
+            .iter()
+            .map(|id| Identifier::Dns(id.to_string()))
+            .collect::<Vec<Identifier>>();
```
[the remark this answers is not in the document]
cpu: Fixed.

tests/pebble.rs
```diff
@@ -294,6 +283,57 @@ impl Environment {
...
+    async fn provision(
+        &self,
+        env: &Environment,
+        identifier: String,
+        challenge: &Challenge,
+        key_auth: &KeyAuthorization,
+    ) -> Result<(), Box<dyn StdError>>;
+}
+
+struct HttpChallengeProvisioner {}
```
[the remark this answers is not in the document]
cpu: Fixed.

> And how about just calling this Http01?

Renamed the structs to `Http01`, `Dns01` and `Alpn01`.

>  (I think I had earlier comments on the naming of the trait, too?)

Oops, sorry I missed that. I've renamed the trait to `AuthorizationMethod` to match your suggestion. I left the trait method `provision` - that felt right to me but I'm open to suggestions if you want to bikeshed it a bit more.

tests/pebble.rs
```diff
@@ -1,3 +1,7 @@
+//! Note: tests in the file are ignored by default because they requires `pebble` and
+//! `pebble-challtestsrv` binaries.
+//!
+//! See documentation for [`PebbleEnvironment`].
```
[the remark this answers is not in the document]
cpu: Fixed.

cpu:
> Probably going to do a round of tweaks after this lands, but I don't know exactly where I am going yet.

Sure, sounds reasonable.

>  In any case, the improved test coverage here is great!

Thanks again for the reviews :-)

PR #78, opened 2025-02-19
Extract Order::poll() method

djc:


src/lib.rs
```diff
@@ -163,6 +165,29 @@ impl Order {
...
+    /// Refresh the order state from the server for `tries` times, waiting `delay` before the
+    /// first attempt and increasing the delay by a factor of 2 for each subsequent attempt.
+    ///
+    /// Yields the [`OrderStatus`] immediately if final or after `tries` attempts.
+    ///
+    /// (Empirically, we've had good results with 5 tries and an initial delay of 250ms.)
+    pub async fn poll(&mut self, mut tries: u8, mut delay: Duration) -> Result<OrderStatus, Error> {
+        loop {
+            sleep(delay).await;
+            let state = self.refresh().await?;
```
djc, line 179: Not sure whether we should keep more trace events in here? Turns out we don't currently have a library-level dependency on tracing here at all, so it's probably orthogonal?

src/lib.rs
```diff
@@ -163,6 +165,29 @@ impl Order {
         self.account.get(&mut self.nonce, challenge_url).await
     }
 
+    /// Poll the order with exponential backoff until in a final state
```
cpu, line 168: nit: missing a `.` on the end of this headline sentence?

src/lib.rs
```diff
@@ -163,6 +165,29 @@ impl Order {
...
+    ///
+    /// Refresh the order state from the server for `tries` times, waiting `delay` before the
+    /// first attempt and increasing the delay by a factor of 2 for each subsequent attempt.
+    ///
+    /// Yields the [`OrderStatus`] immediately if final or after `tries` attempts.
+    ///
+    /// (Empirically, we've had good results with 5 tries and an initial delay of 250ms.)
+    pub async fn poll(&mut self, mut tries: u8, mut delay: Duration) -> Result<OrderStatus, Error> {
+        loop {
+            sleep(delay).await;
```
cpu, line 178: I would have expected this to `self.refresh().await?` once up-front before sleeping. Was always sleeping for the `delay` before the first poll something you wanted to do on purpose?

src/lib.rs
cpu: I feel like it would be useful operationally if it were exposed at a `Trace` level. It feels like the kind of thing you don't care about at all until you're down a debugging rabbit hole and then it might be important (e.g. when trying to figure out rate limit issues).

src/lib.rs
```diff
@@ -163,6 +165,29 @@ impl Order {
         self.account.get(&mut self.nonce, challenge_url).await
     }
 
+    /// Poll the order with exponential backoff until in a final state
+    ///
+    /// Refresh the order state from the server for `tries` times, waiting `delay` before the
+    /// first attempt and increasing the delay by a factor of 2 for each subsequent attempt.
+    ///
+    /// Yields the [`OrderStatus`] immediately if final or after `tries` attempts.
```
cpu, line 173: Do you think it's worth describing what statuses are considered "final"?

cpu (approved):
Nice!

src/lib.rs
djc: Nope! You'll see that libraries which I originated consistently omit the `.` on the first line, on account of this being often more like a sentence fragment then a full sentence.

cpu: Ah! Okie

src/lib.rs
djc: Yes, added that.

src/lib.rs
djc: Fair -- let's postpone until we decide to add more tracing more broadly.

src/lib.rs
cpu: > (but I admit this is a bit corner-case-y).

Probably best to leave the initial sleep since that's most relevant to the vast majority of cases and accept that if you're doing weird stuff you might have to eat one extra sleep. 

I'll resolve the thread :+1:

cpu (approved):

PR #79, opened 2025-02-20
Refactor Pebble authorization method abstraction

djc:
Followup from #77, as promised.

### DIFF
--- a/tests/pebble.rs
+++ b/tests/pebble.rs
@@ -344,16 +344,15 @@ impl Environment {
             .request(
                 Request::builder()
                     .method(Method::POST)
-                    .uri(format!("{}/{}", self.challenge_management_url(), path))
+                    .uri(format!(
+                        "http://[::1]:{}/{}",
+                        self.config.challtestsrv_port, path
+                    ))
                     .header(CONTENT_TYPE, "application/json")
                     .body(Full::from(serde_json::to_vec(body)?))?,
             )
             .await?;
         Ok(())
-    }
-
-    fn challenge_management_url(&self) -> String {
-        format!("http://[::1]:{}", self.config.challtestsrv_port)
     }
 
     fn directory_url(&self) -> String {
--- a/tests/pebble.rs
+++ b/tests/pebble.rs
@@ -50,7 +50,7 @@ async fn http_01() -> Result<(), Box<dyn StdError>> {
 
     // Spawn a Pebble CA and a challenge response server.
     debug!("starting Pebble CA environment");
-    let pebble = EnvironmentConfig::default().run().await?;
+    let pebble = Environment::new(EnvironmentConfig::default()).await?;
 
     // Create a test account with the Pebble CA.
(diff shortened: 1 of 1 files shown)


tests/pebble.rs
```diff
@@ -50,12 +50,10 @@ async fn http_01() -> Result<(), Box<dyn StdError>> {
 
     // Spawn a Pebble CA and a challenge response server.
     debug!("starting Pebble CA environment");
-    let mut pebble = Environment::new().await?;
```
djc: I feel like we're flip-flopping on this -- not sure if you had other reasons than the authorization method repetition to keep these separate?

tests/pebble.rs
```diff
@@ -335,6 +335,19 @@ impl Environment {
         Ok(roots)
     }
 
+    async fn request(&self, path: &str, body: &impl Serialize) -> Result<(), Box<dyn StdError>> {
```
cpu: I think `request` as a name might be too short to convey that it's always using the challtestsrv as a server (vs, say, a request to Pebble, or Pebble's own management interface).

WDYT about `challtestsrv_request`? or `challenge_management_request`? A bit wordy, but at least clear.

tests/pebble.rs
```diff
@@ -130,7 +130,9 @@ struct Environment {
 
 impl Environment {
     /// Create a new [`Environment`] with running Pebble and challenge test servers.
-    async fn new(config: EnvironmentConfig) -> io::Result<Environment> {
+    async fn new() -> io::Result<Environment> {
+        let config = EnvironmentConfig::default();
```
cpu, line 110: This change means it's not possible to customize the `EnvironmentConfig` for a test env above/beyond the default. That seems unnecessarily inflexible, no? It's not much savings to remove the one arg IMO.

tests/pebble.rs
cpu: I don't remember now. In the current moment I agree it makes sense to consolidate :+1:

cpu (approved):
Thanks, this is a nice set of changes. Especially the trait tweaks :+1:

tests/pebble.rs
cpu: What about `Option<EnvironmentConfig>` with an `unwrap_or_default()` as a compromise?

djc: Dropped the commit. Not entirely sure when/why you'd want to customize the `EnvironmentConfig`...

cpu: > Not entirely sure when/why you'd want to customize the EnvironmentConfig...

I was hoping to add a test soon that tests EAB account creation since it's a fairly tricky part of the spec with its own cryptography. That will require a pebble config with `externalAccountBindingRequired: true`. We can't set that in the default, because then we would have to use EAB for every test.

It's a bit "speculative" in that none of the existing tests need it, but I will promise to add a test for it soon :-)

cpu (approved):

PR #81, opened 2025-02-21
Implement RFC 8555 §6.7.1 subproblems

cpu:
When an ACME CA rejects an order submitted by a client it's helpful to be able to indicate _which_ of the order's identifiers are at fault and why. [RFC 8555 §6.7.1](https://www.rfc-editor.org/rfc/rfc8555#section-6.7.1) describes how subproblems within a top-level problem are used for this purpose.

This branch implements support for capturing and rendering problem information containing subproblems based on §6.7.1. Pebble implemented returning subproblems for the case where an identifier in an order is rejected on the basis of CA policy  (https://github.com/letsencrypt/pebble/pull/383). We use this implementation to add a new integration test exercising a problem-with-subproblems response from the test CA.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -131,6 +131,11 @@ pub struct Problem {
     pub detail: Option<String>,
     /// The HTTP status code returned for this response
     pub status: Option<u16>,
+    /// One or more subproblems associated with specific identifiers
+    ///
+    /// See <https://www.rfc-editor.org/rfc/rfc8555#section-6.7.1>
+    #[serde(default)]
+    pub subproblems: Vec<Subproblem>,
 }
 
 impl Problem {
@@ -159,11 +164,55 @@ impl fmt::Display for Problem {
             write!(f, " ({})", r#type)?;
         }
 
+        if !self.subproblems.is_empty() {
+            let count = self.subproblems.len();
+            write!(f, ": {count} subproblems: ")?;
+            for (i, subproblem) in self.subproblems.iter().enumerate() {
+                write!(f, "{subproblem}")?;
+                if i != count - 1 {
+                    f.write_str(", ")?;
+                }
+            }
+        }
+
         Ok(())
     }
 }
 
 impl std::error::Error for Problem {}
+
+/// An RFC 8555 subproblem document contained within a problem returned by the ACME server
+///
+/// See <https://www.rfc-editor.org/rfc/rfc8555#section-6.7.1>
+#[derive(Clone, Debug, Deserialize)]
+pub struct Subproblem {
+    /// The identifier associated with this problem
+    pub identifier: Option<Identifier>,
+    /// One of an enumerated list of problem types
+    ///
+    /// See <https://datatracker.ietf.org/doc/html/rfc8555#section-6.7>
+    pub r#type: Option<String>,
+    /// A human-readable explanation of the problem
+    pub detail: Option<String>,
+}
+
+impl fmt::Display for Subproblem {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(identifier) = &self.identifier {
+            write!(f, r#"for "{identifier}""#)?;
+        }
+
+        if let Some(detail) = &self.detail {
+            write!(f, ": {detail}")?;
+        }
+
+        if let Some(r#type) = &self.r#type {
+            write!(f, " ({})", r#type)?;
+        }
+
+        Ok(())
+    }
+}
 
 #[derive(Debug, Serialize)]
 pub(crate) struct FinalizeRequest {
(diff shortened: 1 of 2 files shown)


src/types.rs
```diff
@@ -131,6 +131,10 @@ pub struct Problem {
     pub detail: Option<String>,
     /// The HTTP status code returned for this response
     pub status: Option<u16>,
+    /// One or more subproblems associated with specific identifiers
+    ///
+    /// See <https://www.rfc-editor.org/rfc/rfc8555#section-6.7.1>
+    pub subproblems: Option<Vec<Subproblem>>,
```
cpu, line 137: Initially I implemented this as a `Option<Vec<Problem>>` and then added an `Optional<Identifier>` field to `Problem`.

In the end I decided that wasn't a great idea because it would allow modelling sub-problems that themselves have sub-problems. Instead it felt like a dedicated `Subproblem` type that didn't admit that possibility into the type-system was preferable.

src/types.rs
```diff
@@ -456,6 +506,14 @@ pub enum Identifier {
     Dns(String),
 }
 
+impl fmt::Display for Identifier {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Dns(domain) => write!(f, "{domain}"),
```
cpu, line 512: My gut instinct here was that the identifier's `String` domain name is sufficient display and that adding a "DNS: " prefix isn't helpful. WDYT?

src/types.rs
```diff
@@ -636,5 +694,81 @@ mod tests {
...
+            second_subproblem.detail,
+            Some(r#"This CA will not issue for "example.net""#.into())
+        );
+
+        let expected_display = "\
+    API error: Some of the identifiers requested were rejected (urn:ietf:params:acme:error:malformed): \
+    2 subproblems: \
+    error for _example.org: Invalid underscore in DNS name \"_example.org\" (urn:ietf:params:acme:error:malformed), \
+    error for example.net: This CA will not issue for \"example.net\" (urn:ietf:params:acme:error:rejectedIdentifier)";
+        assert_eq!(format!("{obj}"), expected_display);
```
cpu, line 772: I'm generally nervous about writing brittle tests based on `Display` impls, but I also wanted to show what a problem with sub-problems renders like. 

WDYT? Keep this or axe it?

src/types.rs
djc: Seems okay to me.

src/types.rs
```diff
@@ -159,12 +163,58 @@ impl fmt::Display for Problem {
...
+#[derive(Clone, Debug, Deserialize)]
+pub struct Subproblem {
+    /// One of an enumerated list of problem types
+    ///
+    /// See <https://datatracker.ietf.org/doc/html/rfc8555#section-6.7>
+    pub r#type: Option<String>,
+    /// A human-readable explanation of the problem
+    pub detail: Option<String>,
+    /// The identifier associated with this problem
+    pub identifier: Option<Identifier>,
```
djc, line 195: Nit: maybe move the `identifier` field to be first (also matches your `Display` impl), as a sort of "key"? (Could even model it as a hashmap, but that's probably a little too opinionated.)

src/types.rs
```diff
@@ -159,12 +163,58 @@ impl fmt::Display for Problem {
...
+}
+
+impl fmt::Display for Subproblem {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("error ")?;
+
+        if let Some(identifier) = &self.identifier {
+            write!(f, "for {identifier}")?;
+        }
```
djc, line 204: How do you feel about dropping `error for `? I think I'd prefer having less "boilerplate" inserted by this impl in favor of emphasizing the content from the server.

src/types.rs
djc: I don't expect us to change this all that often and it is useful for human reviewers, so let's try keeping it for now.

tests/pebble.rs
```diff
@@ -88,6 +88,48 @@ async fn tls_alpn_01() -> Result<(), Box<dyn StdError>> {
...
+async fn forbidden_identifier() -> Result<(), Box<dyn StdError>> {
+    let _ = tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(EnvFilter::from_default_env())
+        .try_init();
+
+    debug!("starting Pebble CA environment");
+    let config = EnvironmentConfig::default();
+    let forbidden_name = config.pebble.domain_blocklist.first().unwrap();
+    let err_result = Environment::new(EnvironmentConfig::default())
```
djc, line 103: Nit: `err_result` seems like a slightly misleading name for something that is `Box<dyn StdError>`? I'd just call it `error` (or maybe even `err`).

tests/pebble.rs
```diff
@@ -88,6 +88,48 @@ async fn tls_alpn_01() -> Result<(), Box<dyn StdError>> {
...
+        .test::<Http01>(&["valid.example.com", forbidden_name])
+        .await
+        .expect_err("issuing for blocked domain name should fail");
+
+    let Error::Api(problem) = *err_result.downcast::<Error>()? else {
+        panic!("unexpected error result");
+    };
+
+    assert_eq!(
+        problem.r#type,
```
djc, line 114: Still pretty irrational, but I'd like to avoid the `.to_string()` calls here if it's not too ugly? I think you could call `.as_ref()` on the left-hand side instead.

djc (approved):

src/types.rs
djc: Seems reasonable. I have a pet peeve that `Option<Vec<_>>` hardly ever seems like a reasonable type, how about expressing this as `Vec<Subproblem>` with `serde(default)`?

cpu: > I have a pet peeve that Option<Vec<_>> hardly ever seems like a reasonable type

I was on the fence about using an empty `Vec` instead and couldn't predict which you'd like better. Switched over :+1:

src/types.rs

> (Could even model it as a hashmap, but that's probably a little too opinionated.)

I thought about that initially but decided against it because I think the spec allows multiple subproblems for a single identifier. Making a map with `Vec` values seemed annoying so I left the type matching the spec's JSON array and figured consumers can adapt to their collection of choice.

src/types.rs
cpu: SGTM.

tests/pebble.rs
cpu: `as_deref()` did the trick, thanks! I fixed both `problem.r#type` comparisons.

tests/pebble.rs

tests/pebble.rs
```diff
@@ -88,6 +88,48 @@ async fn tls_alpn_01() -> Result<(), Box<dyn StdError>> {
         .await
 }
 
+/// Test subproblem handling by trying to issue for a forbidden identifier.
```
djc, line 91: Nit: a trailing dot slipped through (thanks for getting it right on many of the others).

djc (approved):

tests/pebble.rs
cpu: :blush:  Fixed!

PR #84, opened 2025-02-22
Ignore unmaintained *ring* advisory for now

djc:
Let's start here to keep CI passing? And wait for the debris from the advisory to clear...


cpu (approved):
Thanks!

PR #85, opened 2025-02-22
Implement ACME Renewal Information (ARI)

cpu:
This branch implements [ACME Renewal Information](https://www.ietf.org/archive/id/draft-ietf-acme-ari-08.html) based on `draft-08`. While not yet a proposed standard RFC this draft is nearly at that stage, and has been [deployed by Let's Encrypt](https://letsencrypt.org/2023/03/23/improving-resliiency-and-reliability-with-ari/) (and I believe Google as well).

Support brings a few new bits of API surface:

* a new `CertificateIdentifier` type that uniquely identifies a certificate for ARI compatible servers. This requires parsing specific certificate fields to construct, and so without taking a dependency users are expected to provide the relevant `pki_types::Der` data pre-extracted from a certificate.
* for users willing to take an optional `x509-parser` dependency we offer a way to go from `pki_types::CertificateDer<_>` to a `CertificateIdentifier` when the `x509-parser` dependency feature is activated.
* The `DirectoryUrls` type gets a new `renewal_info` field to track the optional ARI endpoint used to signal the ACME server supports ARI.
* When the optional `time` dependency feature is activated, the `Account` struct gains a new `renewal_info()` function that accepts a `CertificateIdentifier` and returns `RenewalInfo` that the caller can use to determine when to replace the identified certificate. This function returns an error if the CA doesn't support ARI.
* The `NewOrder` struct gains a new optional `replaces(cert_id: CertificateIdentifier)` function that can be used to indicate an order is a replacement for a previous order. CAs might use this information to give more generous rate limits, or to know when it's safe to revoke a previously issued certificate due to imposed compliance reasons. Like `Account.renewal_info()` the `Account.new_order()` function will error if provided a `NewOrder` with a `replacement` value if the ACME CA doesn't support ARI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.78"
 thiserror = "2.0.3"
 tokio = { version = "1.22", features = ["time"] }
+x509-parser = { version = "0.17", default-features = false, optional = true }
 
 [dev-dependencies]
 anyhow = "1.0.66"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,3 +59,8 @@ required-features = ["hyper-rustls"]
 [[test]]
 name = "pebble"
 required-features = ["hyper-rustls"]
+
+[package.metadata.docs.rs]
+# all non-default features except fips (cannot build on docs.rs environment)
+features = ["aws-lc-rs", "x509-parser", "time"]
+rustdoc-args = ["--cfg", "docsrs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ ring = { version = "0.17", features = ["std"], optional = true }
 rustls-pki-types = "1.1.0"
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.78"
+time = { version = "0.3", default-features = false, features = ["serde", "parsing"], optional = true }
 thiserror = "2.0.3"
 tokio = { version = "1.22", features = ["time"] }
 x509-parser = { version = "0.17", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "instant-acme"
-version = "0.7.2"
+version = "0.8.0"
 edition = "2021"
 rust-version = "1.70"
 license = "Apache-2.0"
--- a/src/types.rs
+++ b/src/types.rs
@@ -550,6 +550,7 @@ pub enum AuthorizationStatus {
 /// Represent an identifier in an ACME [Order](crate::Order)
 #[allow(missing_docs)]
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[non_exhaustive]
 #[serde(tag = "type", content = "value", rename_all = "camelCase")]
 pub enum Identifier {
     Dns(String),
--- a/src/types.rs
+++ b/src/types.rs
@@ -351,7 +351,16 @@ pub struct OrderState {
 #[serde(rename_all = "camelCase")]
 pub struct NewOrder<'a> {
     /// Identifiers to be included in the order
-    pub identifiers: &'a [Identifier],
+    identifiers: &'a [Identifier],
+}
+
+impl<'a> NewOrder<'a> {
+    /// Prepare to create a new order for the given identifiers
+    ///
+    /// To be passed into [Account::new_order()](crate::Account::new_order()).
+    pub fn new(identifiers: &'a [Identifier]) -> Self {
+        Self { identifiers }
+    }
 }
 
 /// Payload for a certificate revocation request
--- a/src/types.rs
+++ b/src/types.rs
@@ -15,6 +15,7 @@ use crate::crypto::{self, KeyPair};
 
 /// Error type for instant-acme
 #[derive(Debug, Error)]
+#[non_exhaustive]
 pub enum Error {
     /// An JSON problem as returned by the ACME server
     ///
--- a/src/types.rs
+++ b/src/types.rs
@@ -9,6 +9,10 @@ use serde::de::{self, DeserializeOwned};
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+#[cfg(feature = "x509-parser")]
+use x509_parser::extensions::ParsedExtension;
+#[cfg(feature = "x509-parser")]
+use x509_parser::parse_x509_certificate;
 
 use crate::BytesResponse;
 use crate::crypto::{self, KeyPair};
@@ -708,6 +712,37 @@ impl Serialize for CertificateIdentifier<'_> {
     }
 }
 
+#[cfg(feature = "x509-parser")]
+impl<'a> TryFrom<&'a CertificateDer<'_>> for CertificateIdentifier<'_> {
+    type Error = String;
+
+    fn try_from(cert: &'a CertificateDer) -> Result<Self, Self::Error> {
+        let (_, parsed_cert) = parse_x509_certificate(cert.as_ref())
+            .map_err(|e| format!("failed to parse certificate: {e}"))?;
+
+        let Some(authority_key_identifier) =
+            parsed_cert
+                .iter_extensions()
+                .find_map(|ext| match ext.parsed_extension() {
+                    ParsedExtension::AuthorityKeyIdentifier(aki_ext) => aki_ext
+                        .key_identifier
+                        .as_ref()
+                        .map(|aki| Der::from_slice(aki.0)),
+                    _ => None,
+                })
+        else {
+            return Err(
+                "certificate does not contain an Authority Key Identifier (AKI) extension".into(),
+            );
+        };
+
+        Ok(Self::new(
+            authority_key_identifier,
+            Der::from_slice(parsed_cert.tbs_certificate.raw_serial()),
+        ))
+    }
+}
+
 impl fmt::Display for CertificateIdentifier<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_str(&self.authority_key_identifier)?;
@@ -740,6 +775,12 @@ pub(crate) struct Empty {}
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "x509-parser")]
+    use rcgen::{
+        BasicConstraints, CertificateParams, DistinguishedName, IsCa, KeyIdMethod, KeyPair,
+        SerialNumber,
+    };
+
     use super::*;
 
     // https://datatracker.ietf.org/doc/html/rfc8555#section-7.4
@@ -950,5 +991,34 @@ mod tests {
 
         let serialized = serde_json::to_string(&cert_id).unwrap();
         assert_eq!(serialized, r#""aYhba4dGQEHhs3uEe6CuLN4ByNQ.AIdlQyE""#);
+    }
+
+    #[cfg(feature = "x509-parser")]
+    #[test]
+    fn encoded_certificate_identifier_from_cert() {
+        // Generate a CA key_pair and self-signed cert with a specific subject key identifier.
+        let ca_key_id = vec![0xC0, 0xFF, 0xEE];
+        let ca_key = KeyPair::generate().unwrap();
+        let mut ca_params = CertificateParams::default();
+        ca_params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+        ca_params.key_identifier_method = KeyIdMethod::PreSpecified(ca_key_id);
+        let ca_cert = ca_params.self_signed(&ca_key).unwrap();
+
+        // Generate an end entity certificate issued by the CA, with a specific serial number
+        // and an AKI extension.
+        let ee_key = KeyPair::generate().unwrap();
+        let ee_serial = [0xCA, 0xFE];
+        let mut ee_params = CertificateParams::new(["example.com".to_owned()]).unwrap();
+        ee_params.distinguished_name = DistinguishedName::new();
+        ee_params.serial_number = Some(SerialNumber::from_slice(ee_serial.as_slice()));
+        ee_params.use_authority_key_identifier_extension = true;
+        let ee_cert = ee_params.signed_by(&ee_key, &ca_cert, &ca_key).unwrap();
+
+        // Extract the AKI and serial number from the EE certificate and create an encoded
+        // certificate identifier.
+        let encoded = CertificateIdentifier::try_from(ee_cert.der()).unwrap();
+
+        // We should arrive at the expected encoded certificate identifier.
+        assert_eq!(format!("{encoded}"), "wP_u.AMr-");
     }
 }
--- a/src/types.rs
+++ b/src/types.rs
@@ -342,6 +342,10 @@ pub struct OrderState {
     pub finalize: String,
     /// The certificate URL, which becomes available after finalization
     pub certificate: Option<String>,
+    /// The certificate that this order is replacing, if any
+    #[serde(deserialize_with = "deserialize_static_certificate_identifier")]
+    #[serde(default)]
+    pub replaces: Option<CertificateIdentifier<'static>>,
 }
 
 /// Input data for [Order](crate::Order) creation
@@ -350,6 +354,8 @@ pub struct OrderState {
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct NewOrder<'a> {
+    /// The [`CertificateIdentifier`] of a previously issued certificate being replaced by the order
+    pub(crate) replaces: Option<CertificateIdentifier<'a>>,
     /// Identifiers to be included in the order
     identifiers: &'a [Identifier],
 }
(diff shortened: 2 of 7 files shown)


src/types.rs
```diff
@@ -335,6 +340,8 @@ pub struct OrderState {
     pub finalize: String,
     /// The certificate URL, which becomes available after finalization
     pub certificate: Option<String>,
+    /// The certificate that this order is replacing
+    pub replaces: Option<EncodedCertificateIdentifier>,
```
cpu, line 344: FWIW: calling out that this will be a breaking change but I haven't bumped the crate version yet in case you want to approach it a different way.

cpu:
I think there might be some kind of a flaky race condition with the Pebble tests that also needs investigating. I'm seeing the occasional spurious error from an order's identifiers not matching the CSR identifiers :thinking:

djc:
Why is it useful for the identifier type to live in pki-types? What would the downside be if it living in this crate?

src/types.rs
```diff
@@ -586,6 +586,23 @@ impl ZeroSsl {
     }
 }
 
+/// A unique certificate identifier for the ACME renewal information (ARI) extension
+///
+/// See <https://www.ietf.org/archive/id/draft-ietf-acme-ari-07.html#section-4.1> for
+/// more information.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct EncodedCertificateIdentifier(pub(crate) String);
```
djc, line 612: Should we make this contain a `Cow<'a, str>` instead?

src/types.rs
djc: Should we make this `#[non_exhaustive]` while we're at it? I think this is pretty much an output-only type, so I think that could make sense?

src/types.rs
```diff
@@ -348,6 +350,17 @@ pub struct OrderState {
...
+    ///
+    /// Some ACME servers may give preferential rate limits to orders that replace
+    /// existing certificates, or use this information to determine when it is safe
+    /// to revoke a certificate affected by a compliance incident.
+    ///
+    /// When provided, at least one of the `identifiers` for the new order must have been
+    /// present in the certificate being replaced. If the ACME CA does not support the
+    /// ACME renewal information (ARI) extension, the [Account::new_order()] method will
+    /// return an error.
+    pub replaces: Option<EncodedCertificateIdentifier>,
```
djc, line 365: How do you feel about inserting a commit before this that changes it from having `pub` fields to having a `new(identifiers: &'a [Identifier])`, and then inserting this as a builder method `.replaces()`? Many/most existing callers will not care to specify `replaces: None`.

src/types.rs
```diff
@@ -619,6 +621,50 @@ impl From<AcmeRenewalInfoCertIdentifier<'_>> for EncodedCertificateIdentifier {
     }
 }
 
+#[cfg(feature = "x509-parser")]
+impl<'a> TryFrom<&'a [u8]> for EncodedCertificateIdentifier {
```
djc, line 625: I think I'd prefer to skip this impl in favor of only the one for `&CertificateDer`? IIRC the conversion from `&[u8]` to `CertificateDer` is ~free anyway, and that would give us some more (explicit) type safety.

src/types.rs
```diff
@@ -619,6 +621,50 @@ impl From<AcmeRenewalInfoCertIdentifier<'_>> for EncodedCertificateIdentifier {
...
+    type Error = String;
+
+    fn try_from(certificate_der: &'a [u8]) -> Result<Self, Self::Error> {
+        let (_, parsed_cert) = x509_parser::parse_x509_certificate(certificate_der)
```
djc, line 629: Nit: let's import `parse_x509_certificate()` and `ParsedExtension` directly?

src/types.rs
```diff
@@ -665,6 +665,30 @@ impl<'a> TryFrom<&'a CertificateDer<'_>> for EncodedCertificateIdentifier {
...
+/// A suggested renewal window for a certificate
+///
+/// See <https://www.ietf.org/archive/id/draft-ietf-acme-ari-07.html#section-4.2>
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SuggestedWindow {
+    /// The RFC3339 formatted start date/time of the suggested renewal window
+    pub start: String,
+    /// The RFC3339 formatted end date/time of the suggested renewal window
+    pub end: String,
```
djc, line 689: I think these fields should be `DateTime<Utc>` directly. Maybe that means that `renewal_info()` is `cfg(feature = "chrono")`?

djc (commented):
Looking pretty good!

cpu:
> Why is it useful for the identifier type to live in pki-types? What would the downside be if it living in this crate?

I flip flopped a bit. Initially I was thinking there was a reason to involve `rcgen`, and so it seemed like a helpful bridge between `instant-acme` and `rcgen`. In practice I think it made more sense to do the heavy lifting here. I closed the pki-types PR and will rebase with a similar type defined here instead.

> I think there might be some kind of a flaky race condition with the Pebble tests that also needs investigating. I'm seeing the occasional spurious error from an order's identifiers not matching the CSR identifiers 🤔

I figured this out locally. It's authz reuse & a logic bug in the test finalization logic carried forward from the `provision.rs` example. I'll fix shortly.

src/types.rs
cpu: Good idea. Done.

I also added some commits later on to switch the Pebble common test logic to accept a `&NewOrder` instead of `&[&'static str]` identifiers. This let those tests avoid unnecessary `None` arguments in a similar manner.

src/types.rs

src/types.rs

src/types.rs

cpu:
I rebased this on top of https://github.com/djc/instant-acme/pull/86 and implemented the review feedback so far.

src/types.rs
```diff
@@ -1,18 +1,24 @@
 use std::fmt;
 
+use crate::BytesResponse;
+use crate::crypto::{self, KeyPair};
```
cpu, line 4: oops, biffed the import order in my rebase. I'll fix this.

Cargo.toml
```diff
@@ -23,6 +23,7 @@ async-trait = "0.1"
 aws-lc-rs = { version = "1.8.0", optional = true }
 base64 = "0.22"
 bytes = "1"
+chrono = { version = "0.4", features = ["serde"], optional = true }
```
cpu, line 26: ah, this commit was meant to be squashed into its parent. Will fix.

cpu:
> This branch has conflicts that must be resolved

Will fix this & the other nits from self-review shortly. Out of juice for today

djc:
Yup, won't take a look at it until tomorrow either.

src/types.rs
djc: Not really -- just wondering if people might have pre-encoded certificate identifiers that they might want to splice in? Like, if you're storing a bunch of certs in your database (like we used to do at my old startup), it wouldn't be crazy to store these directly to make querying ARI easier, and in that case it feels like low-effort future-proofing to enable borrowing this down the line.

Also, how important is the `Encoded` prefix here? In my quest to strive for minimal entropy in type/variable names, is there another kind/encoding of certificate identifier that we need to disambiguate from?

Digging into this a little bit more, I noticed that the spec also doesn't really define the concept of a certificate identifier (maybe good spec feedback?). I think my ideal representation might be more like this:

```rust
struct CertificateId {
    authority_key_identifier: Cow<'a, str>,
    serial: Cow<'a, str>,
}
```

(Perhaps even with `pub` fields?) WDYT?

tests/pebble.rs
```diff
@@ -184,6 +184,17 @@ fn try_tracing_init() {
         .try_init();
 }
 
+fn dns_identifiers<I, T>(dns_names: I) -> Vec<Identifier>
```
djc, line 242: Nits: move this down below its user? Also use `impl Trait` to remove the indirection, I think we could write `names: impl IntoIterator<Item = impl ToString>`?

Cargo.toml
djc: Hmm, also on consideration, rcgen and yasna depend on time instead of chrono today. I wonder if that mean we should use time instead of chrono here?

djc: Also, make sure that we have `doc_auto_cfg` set up so `renewal_info()` will be discoverable in the docs even when `cfg`ed out by default.

src/types.rs
cpu: Fixed.

src/types.rs
cpu: > Not really -- just wondering if people might have pre-encoded certificate identifiers that they might want to splice in?

Good point :+1: 

>  Also, how important is the Encoded prefix here? 

I removed the prefix. I think `CertificateIdentifier` is clear enough, especially now that it doesn't hold the encoded & concatinated form directly.

> Digging into this a little bit more, I noticed that the spec also doesn't really define the concept of a certificate identifier (maybe good spec feedback?).

Sorry, still a bit too burnt out on IETF to want to jump in with feedback outside of work contexts. Even small improvements like this have a way of turning into long drawn-out & exhausting conversations. I agree it would be nicer if the spec gave it an explicit name.

> I think my ideal representation might be more like this:
> ...
> (Perhaps even with pub fields?) WDYT?

I like it a lot! I went ahead and implemented this. It took some work to get the ser/der stuff working nicely but I'm pretty happy with how it turned out.

Cargo.toml
cpu: >  I wonder if that mean we should use time instead of chrono here?

Fair point. I switched to using `time`.

> Also, make sure that we have doc_auto_cfg set up so renewal_info() will be discoverable in the docs even when cfged out by default.

I tacked a commit on at the end to try and make this work. Like other repos we have to avoid the `fips` feature but otherwise it seems workable.

tests/pebble.rs
cpu: Done x2.

cpu:
> cpu marked this pull request as ready for review now

I think this is in pretty good shape now. I'd like to hold merging it until I have a chance to verify it works with LE's staging server but I should be able to do that in the next few days.

src/types.rs
```diff
@@ -586,6 +589,106 @@ impl ZeroSsl {
...
+}
+
+impl<'de: 'a, 'a> Deserialize<'de> for CertificateIdentifier<'a> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct CertificateIdentifierVisitor<'de> {
+            marker: PhantomData<CertificateIdentifier<'de>>,
+        }
```
djc, line 692: One of my pet peeves is that custom `Deserialize` implementations are very verbose. I've found that a good approach to improve that is to define an intermediate type and implement the `Deserialize` implementation in terms of a processing step on top of a derived `Deserialize` implementation. Sort of like this:

```rust
#[derive(Deserialize)]
struct CertificateIdentifierStr<'a>(&'a str);

let s = CertificateIdentifierStr::deserialize(deserializer)?;
let Some((aki, serial)) = value.split_once('.') else {
     ..
};

if serial.contains('.') {
     return Err(..);
}

Ok(CertificateIdentifier { .. })
```

src/types.rs
```diff
@@ -586,6 +589,106 @@ impl ZeroSsl {
...
+
+        deserializer.deserialize_str(CertificateIdentifierVisitor {
+            marker: PhantomData,
+        })
+    }
+}
+
+impl Serialize for CertificateIdentifier<'_> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
djc, line 728: Nit: same here with the inline bound?

src/types.rs
```diff
@@ -343,24 +343,50 @@ pub struct OrderState {
...
-pub struct NewOrder<'a> {
+pub struct NewOrder<'a, 'b> {
+    /// The [`CertificateIdentifier`] of a previously issued certificate being replaced by the order
+    pub(crate) replaces: Option<CertificateIdentifier<'b>>,
     /// Identifiers to be included in the order
     identifiers: &'a [Identifier],
 }
 
-impl<'a> NewOrder<'a> {
+impl<'a, 'b> NewOrder<'a, 'b> {
```
djc, line 370: Do we actually need a second lifetime for this? I'd expect we can just reuse `'a` for this.

src/types.rs
```diff
@@ -703,6 +729,19 @@ impl fmt::Display for CertificateIdentifier<'_> {
...
+    deserializer: D,
+) -> Result<Option<CertificateIdentifier<'static>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let opt_cert_id = Option::<CertificateIdentifier<'_>>::deserialize(deserializer)?;
+    Ok(opt_cert_id.map(move |cert_id| CertificateIdentifier {
+        authority_key_identifier: Cow::Owned(cert_id.authority_key_identifier.into_owned()),
+        serial: Cow::Owned(cert_id.serial.into_owned()),
+    }))
```
djc, line 814: Maybe reformulate more like this?

```rust
let Some(cert_id) = Option::<CertificateIdentifier<'_>>::deserialize(deserializer)? else {
    return Ok(None);
};

Ok(CertificateIdentifier { .. })
```

Also I'd suggest outlining the `into_owned()` into a `CertificateIdentifier::into_owned()` so as to make this code accessible to outside callers as well.

djc (commented):
Also, don't forget to add a bullet point in the README!

src/types.rs
```diff
@@ -586,6 +589,106 @@ impl ZeroSsl {
...
+            authority_key_identifier: BASE64_URL_SAFE_NO_PAD
+                .encode(authority_key_identifier)
+                .into(),
+            serial: BASE64_URL_SAFE_NO_PAD.encode(serial).into(),
+        }
+    }
+}
+
+impl<'de: 'a, 'a> Deserialize<'de> for CertificateIdentifier<'a> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
```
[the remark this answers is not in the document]
cpu: Done (throughout)

src/types.rs
cpu: Ohhh that's so much nicer. Thank you. I think I've only had to do this ~2 or so times and didn't think about this approach.

Reworked!

src/types.rs

src/types.rs
cpu: Switched to one lifetime.

src/types.rs
cpu: Done x2. I put the `into_owned()` addition in its own commit up-front.

cpu:
> Also, don't forget to update the README -- I think this is worth a feature bullet point.

Done! I also bumped the crate version to reflect the breaking changes. Moving on to testing with a real server.

src/types.rs
```diff
@@ -582,6 +640,164 @@ impl ZeroSsl {
...
+                .encode(authority_key_identifier)
+                .into(),
+            serial: BASE64_URL_SAFE_NO_PAD.encode(serial).into(),
+        }
+    }
+
+    /// Convert the `CertificateIdentifier` into an owned version with a static lifetime
+    pub fn into_owned(self) -> CertificateIdentifier<'static> {
+        CertificateIdentifier {
+            authority_key_identifier: Cow::Owned(self.authority_key_identifier.into_owned()),
```
djc, line 686: I think you can call `Cow::to_owned()` to make this a little neater.

djc (approved):

src/types.rs
cpu: Writing:
```rust
    pub fn into_owned(self) -> CertificateIdentifier<'static> {
        CertificateIdentifier {
            authority_key_identifier: Cow::to_owned(&self.authority_key_identifier),
            serial:  Cow::to_owned(&self.serial),
        }
    }
```

gives a lifetime error. Am I missing something?

cpu:
> Moving on to testing with a real server.

This ended up being a bit of a rabbit hole and ultimately resulted in one code change. Fetching renewal info works great, but creating replacement orders was throwing an error because the reflected order doesn't have a matching `replaces` field.

Previously in `new_order()` I was following the spec text where it says:

> If the Server accepts a new-order request with a "replaces" field, it MUST reflect that field in the response and in subsequent requests for the corresponding Order object.

If we sent `replaces`, I asserted the server echoed the right `replaces` value. That works great with Pebble :+1: 

With Let's Encrypt staging and production I never get back a reflected `replaces` in an echoed order object. I thought at first this was related to "order re-use" since Pebble doesn't implement this, but even after I implemented de-activating an order (PR forthcoming) a brand new order created with a `replaces` in the `NewOrder` req still had no `replaces` echoed.

From my cursory look at Boulder, there doesn't seem to be a place to set that field at all. The internal [`corepb.Order` protobuf message](https://github.com/letsencrypt/boulder/blob/10d9ef9af7095298d22582e6fd519a2672e69c04/core/proto/core.proto#L110-L128) doesn't have the information, and neither does the rendered [`wfe2.orderJSON`](https://github.com/letsencrypt/boulder/blob/10d9ef9af7095298d22582e6fd519a2672e69c04/wfe2/wfe.go#L1915-L1924) type marshalled into the response.

I'll file an upstream issue but for now I've relaxed the check in 9d7563b5e37b8c811a3389fe470b51554ccd5dda and suggest we can merge this as-is. @djc WDYT?

cpu:
> I'll file an upstream issue

https://github.com/letsencrypt/boulder/issues/8034

src/types.rs
djc: Call it as a method on the field?

djc:
> I'll file an upstream issue but for now I've relaxed the check in 9d7563b and suggest we can merge this as-is. @djc WDYT?

LGTM!

src/types.rs
cpu: You mean like this?

```rust
        CertificateIdentifier {
            authority_key_identifier: self.authority_key_identifier.to_owned(),
            serial: self.serial.to_owned(),
        }
``` 

Same result :thinking:

cpu: I'm going to leave it as is, but if you want to take a crack at cleaning this up I'd be interested to see what I'm missing :laughing:

djc: Hmm, I think I had too high expectations of the `impl<T> ToOwned for Cow`...

PR #86, opened 2025-02-23
Fix authorization reuse in provision.rs and the Pebble integration tests

cpu:
Previously both the provision example and the Pebble integration test logic created a new order for a set of input identifiers, iterated the authorizations for the resulting order, and then pulled out the identifier/challenge for each **pending** status authz. After POSTing each authz's challenge URL the associated challenge identifiers were used for the CSR sent to finalize the order.

Since the logic above only processes pending authz's it would skip any that are already **valid** (e.g. from previous orders). This can end up producing a CSR that has fewer names than the initial order, which provokes an error during finalization when the CA rejects the CSR.

The solution is simple: don't tie the CSR identifiers to the challenges at all, use the same list of identifiers as was used to create the order. Everything else remains the same.

This was manifesting as integration test flakes in https://github.com/djc/instant-acme/pull/85 where we added a test that made multiple orders one after another. Pebble re-uses valid authzs 50% of the time by default, so there was a 50% chance that an incorrect CSR would be submitted, breaking the test.

### DIFF
--- a/tests/pebble.rs
+++ b/tests/pebble.rs
@@ -286,7 +286,6 @@ impl Environment {
 
         let authorizations = order.authorizations().await?;
         let mut challenges = Vec::with_capacity(authorizations.len());
-        let mut names = Vec::with_capacity(authorizations.len());
 
         // Collect up the relevant challenges, provisioning the expected responses as we go.
         for authz in &authorizations {
@@ -307,12 +306,11 @@ impl Environment {
             let key_authz = order.key_authorization(challenge);
             self.request_challenge::<A>(identifier, challenge, &key_authz)
                 .await?;
-            challenges.push((identifier, &challenge.url));
-            names.push(identifier.clone());
+            challenges.push(&challenge.url);
         }
 
         // Tell the CA we have provisioned the response for each challenge.
-        for (_, url) in &challenges {
+        for url in &challenges {
             debug!(challenge_url = url, "marking challenge ready");
             order.set_challenge_ready(url).await?;
         }
@@ -324,7 +322,7 @@ impl Environment {
         }
 
         // Issue a certificate for the names and test the chain validates to the issuer root.
-        let cert_chain = self.certificate(&mut order, &names).await?;
+        let cert_chain = self.certificate(&mut order, names).await?;
 
         // Split off and parse the EE cert, save the intermediates that follow.
         let (ee_cert, intermediates) = cert_chain.split_first().unwrap();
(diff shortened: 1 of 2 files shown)


tests/pebble.rs
```diff
@@ -355,12 +353,17 @@ impl Environment {
...
+        identifiers: &[&str],
     ) -> Result<Vec<CertificateDer<'static>>, Box<dyn StdError>> {
         info!(?identifiers, order_url = order.url(), "issuing certificate");
 
         // Create a CSR for the identifiers corresponding to the order.
-        let mut params = CertificateParams::new(identifiers.to_owned())?;
+        let mut params = CertificateParams::new(
+            identifiers
+                .iter()
+                .map(|s| (*s).to_owned())
```
djc, line 395: Nit: `|&s| s.to_owned()`?

tests/pebble.rs
```diff
@@ -355,12 +353,17 @@ impl Environment {
...
+                .collect::<Vec<_>>(),
```
djc, line 396: I presume you've tested that the turbofish is necessary here?

tests/pebble.rs
```diff
@@ -557,6 +562,7 @@ struct EnvironmentConfig {
     dns_port: u16,
     challtestsrv_port: u16,
     eab_key: Option<ExternalAccountKey>,
+    reuse_all_authz: bool,
```
djc, line 591: How about we make this a more direct match to the Pebble config:

```rust
/// Percentage of authorizations that can be reused [maybe there's upstream docs that describe this better?]
authz_reuse: u8,
```

djc (approved):
Nice work!

tests/pebble.rs
cpu: Yeah, unfortunately necessary.

<details>
<summary>With a bare `collect()`:</summary>

```rust
error[E0283]: type annotations needed
   --> tests/pebble.rs:365:18
    |
361 |         let mut params = CertificateParams::new(
    |                          ---------------------- required by a bound introduced by this call
...
365 |                 .collect(),
    |                  ^^^^^^^ cannot infer type of the type parameter `B` declared on the method `collect`
    |
    = note: cannot satisfy `_: std::convert::Into<std::vec::Vec<std::string::String>>`
note: required by a bound in `rcgen::CertificateParams::new`
   --> /home/daniel/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rcgen-0.13.2/src/certificate.rs:122:37
    |
122 |     pub fn new(subject_alt_names: impl Into<Vec<String>>) -> Result<Self, Error> {
    |                                        ^^^^^^^^^^^^^^^^^ required by this bound in `CertificateParams::new`
help: consider specifying the generic argument
    |
365 |                 .collect::<Vec<_>>(),
    |                         ++++++++++
```
</details>

tests/pebble.rs

tests/pebble.rs
cpu: Done and linked to [the upstream docs](https://github.com/letsencrypt/pebble?tab=readme-ov-file#valid-authorization-reuse).

PR #89, opened 2025-03-02
Add authorization and account deactivation

cpu:
Add support for deactivating authorizations, and ACME accounts. 

Additionally add a wildcard identifier to the DNS-01 integration test and fix the issued certificate validation logic to support this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,10 +70,15 @@ impl Order {
     /// After the challenges have been set up, check the [`Order::state()`] to see
     /// if the order is ready to be finalized (or becomes invalid). Once it is
     /// ready, call `Order::finalize()` to get the certificate.
-    pub async fn authorizations(&mut self) -> Result<Vec<AuthorizationState>, Error> {
+    pub async fn authorizations(&mut self) -> Result<Vec<Authorization>, Error> {
         let mut authorizations = Vec::with_capacity(self.state.authorizations.len());
         for url in &self.state.authorizations {
-            authorizations.push(self.account.get(&mut self.nonce, url).await?);
+            authorizations.push(Authorization {
+                account: self.account.clone(),
+                nonce: None,
+                url: url.clone(),
+                state: self.account.get(&mut self.nonce, url).await?,
+            });
         }
         Ok(authorizations)
     }
@@ -216,6 +221,45 @@ impl Order {
     }
 
     /// Get the URL of the order
+    pub fn url(&self) -> &str {
+        &self.url
+    }
+}
+
+/// An ACME authorization as described in RFC 8555 (section 7.1.4)
+///
+/// Authorizations are retrieved from an associated [`Order`] by calling
+/// [`Order::authorizations()`].
+///
+/// <https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.3>
+pub struct Authorization {
+    account: Arc<AccountInner>,
+    nonce: Option<String>,
+    url: String,
+    state: AuthorizationState,
+}
+
+impl Authorization {
+    /// Refresh the current state of the authorization
+    pub async fn refresh(&mut self) -> Result<&AuthorizationState, Error> {
+        let rsp = self
+            .account
+            .post(None::<&Empty>, self.nonce.take(), &self.url)
+            .await?;
+
+        self.nonce = nonce_from_response(&rsp);
+        self.state = Problem::check::<AuthorizationState>(rsp).await?;
+        Ok(&self.state)
+    }
+
+    /// Get the [`AuthorizationState`] of the authorization
+    ///
+    /// Call `refresh()` to get the latest state from the server.
+    pub fn state(&self) -> &AuthorizationState {
+        &self.state
+    }
+
+    /// Get the URL of the authorization
     pub fn url(&self) -> &str {
         &self.url
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -584,6 +584,35 @@ impl Account {
         Problem::check::<RenewalInfo>(rsp).await
     }
 
+    /// Deactivate the account with the ACME server
+    ///
+    /// This is useful when you want to cancel an account with the ACME server
+    /// because you don't intend to use it further, or because the account key was
+    /// compromised.
+    ///
+    /// After this point no further operations can be performed with the account.
+    /// Any existing orders or authorizations created with the ACME server will be
+    /// invalidated.
+    pub async fn deactivate(self) -> Result<(), Error> {
+        #[derive(Serialize)]
+        struct DeactivateRequest<'a> {
+            status: &'a str,
+        }
+
+        let _ = self
+            .inner
+            .post(
+                Some(&DeactivateRequest {
+                    status: "deactivated",
+                }),
+                None,
+                self.id(),
+            )
+            .await?;
+
+        Ok(())
+    }
+
     /// Get the account ID
     pub fn id(&self) -> &str {
         &self.inner.id
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,6 +252,52 @@ impl Authorization {
         Ok(&self.state)
     }
 
+    /// Deactivate a pending or valid authorization
+    ///
+    /// Returns the updated [`AuthorizationState`] if the deactivation was successful.
+    /// If the authorization was not pending or valid, an error is returned.
+    ///
+    /// Once deactivated the authorization and associated challenges can not be updated
+    /// further.
+    ///
+    /// This is useful when you want to cancel a pending authorization attempt you wish
+    /// to abandon, or if you wish to revoke valid authorization for an identifier to
+    /// force future uses of the identifier by the same ACME account to require
+    /// re-verification with fresh authorizations/challenges.
+    pub async fn deactivate(&mut self) -> Result<&AuthorizationState, Error> {
+        if !matches!(
+            self.state.status,
+            AuthorizationStatus::Pending | AuthorizationStatus::Valid
+        ) {
+            return Err(Error::Other("authorization not pending or valid".into()));
+        }
+
+        #[derive(Serialize)]
+        struct DeactivateRequest {
+            status: AuthorizationStatus,
+        }
+
+        let rsp = self
+            .account
+            .post(
+                Some(&DeactivateRequest {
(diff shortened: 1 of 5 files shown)


tests/pebble.rs
```diff
@@ -230,6 +233,69 @@ async fn replacement_order() -> Result<(), Box<dyn StdError>> {
...
+    }
+
+    // With all authz's deactivated, the order should be status == Invalid
+    // However, until https://github.com/letsencrypt/pebble/pull/301 is merged Pebble returns
+    // an unspecified "deactivated" state we reject as a JSON decoding error about an unexpected
+    // status.
+    // TODO(@cpu): when Pebble is fixed, replace with an assertion on order status.
+    let err = order.deactivate().await.unwrap_err();
+    assert!(matches!(err, Error::Json(_)));
+    assert!(err.to_string().contains("deactivated"));
```
cpu, line 260: This is specific to Pebble, Boulder has already been fixed. I'll ping the maintainers during business hours to see about getting the fix merged & a release cut.

djc:
I'm having trouble wrapping my head around the motivation for authz deactivation? Why is it useful? Is there no way to cancel an order in its entirety?

cpu:
> I'm having trouble wrapping my head around the motivation for authz deactivation? Why is it useful? Is there no way to cancel an order in its entirety?

Correct, AFAIK there's no way to cancel an order other than deactivating the authorizations, or POSTing one of the challenges and failing it.

Operationally I think there's value in being able to leave your ACME account in a state post-issuance where you have a certificate for some identifiers in hand, but also know that creating a new certificate for those identifiers would require another successful proof of domain control. It takes some threat model contorting to make this an important consideration (_maybe you interact with another microservice that gives you a short-lived token to update the DNS zone  for proof of control and don't want the account's capabilities to exceed that token's lifetime_) but I think it's reasonable-ish (authzs can have a lifetime on the order of days to weeks).

More concretely I found it useful while testing because I thought a bug I was observing might be specific to order-reuse and so needed to ensure the CA gave me a fresh order by deactivating the valid authz's from the previous one.

I don't feel strongly if you think the utility is too speculative. WDYT?

src/types.rs
```diff
@@ -552,6 +552,7 @@ pub enum AuthorizationStatus {
     Invalid,
     Revoked,
     Expired,
+    Deactivated,
```
djc, line 555: Nit: I don't think this is worth its own commit.

src/lib.rs
```diff
@@ -168,6 +168,56 @@ impl Order {
...
+    /// Deactivate the order's pending or valid authorizations
+    ///
+    /// Returns the updated status of the order post-deactivation.
+    ///
+    /// This is useful when you want to cancel a pending order you wish to abandon, or to revoke
+    /// valid authorizations for the order's identifiers to prevent reuse.
+    ///
+    /// Once deactivated the order and associated resources (authorizations, challenges, etc) can
+    /// not be updated further.
+    pub async fn deactivate(&mut self) -> Result<&OrderState, Error> {
```
djc, line 180: This feels like a comparatively fat abstraction. How do you feel about attaching something like this to the `Authorization` type directly? `Order` already has an `authorizations()` method... You'd have to `refresh()` by hand.

src/lib.rs
```diff
@@ -168,6 +168,56 @@ impl Order {
...
+        #[derive(Serialize)]
+        struct DeactivateRequest<'a> {
+            status: &'a str,
```
djc, line 183: Would it make semantic sense to have this take an `AuthorizationStatus`?

tests/pebble.rs
djc: Ohh, you have a long history with this stuff. 😄

tests/pebble.rs
```diff
@@ -264,6 +264,35 @@ async fn order_deactivate() -> Result<(), Box<dyn StdError>> {
     Ok(())
 }
 
+/// Test account deactivation
+#[tokio::test]
+#[ignore]
+async fn acct_deactivate() -> Result<(), Box<dyn StdError>> {
```
djc, line 273: Nit: don't feel like abbreviating `account` in the test name is worth it.

djc (commented):
Okay, I think I'm willing to take some form of this.

src/lib.rs
djc: How about doing something like this:

```rust
struct Order {
     authorizations: Vec<Authorization>,
}

struct Authorization {
     pub url: String,
     pub state: Option<AuthorizationState>,
     // maybe:
     account: Arc<AccountInner>,
}

impl Authorization {
     pub async fn refresh(&mut self) -> Result<&AuthorizaionState, Error> { .. }

     pub async fn deactivate(&mut self) -> Result<(), Error> { .. }
}
```

What do you think?

cpu: That seems like a nice solution. I'll give it a go :+1:

tests/pebble.rs
cpu: The fix was merged upstream after a bump so this workaround fell away.

There's no release yet so I added a commit up-front to build Pebble from src in CI. The time difference is minimal and we can revert once there's a new release w/ updated binaries.

src/types.rs
cpu: Folded into the commit that adds deactivation.

src/lib.rs
cpu: Fixed. I added a commit implementing serialize for `AuthorizationStatus` into a separate commit. I can fold that into the one adding deactivation if preferred.

tests/pebble.rs
cpu: Fixed.

src/lib.rs
djc: Yes, would prefer squashing.

src/lib.rs
cpu: Here's a first pass: 319be2335821718e38fd51ac125de8106e91ee76

I more or less matched the approach `Order` and `OrderState` use (e.g. accessors vs `pub` fields). It feels to me like the existing behaviour of `Order::authorizations()` where you got back the state immediately for each authz made sense, so I implemented `Authorization` to hold non-optional `AuthorizationState` instead of `Option<AuthorizationState`. If you think it's better to populate lazily and have `authorizations()` return `state: None` I can rework.

With the nonce stuff factored in it didn't seem like `RefCell` added much, the lifetimes in use seemed OK, but I'm open to further iteration if you have suggestions!

src/lib.rs
cpu: Squashed :foot:

src/lib.rs
```diff
@@ -252,6 +252,52 @@ impl Authorization {
...
+        if !matches!(
+            self.state.status,
+            AuthorizationStatus::Pending | AuthorizationStatus::Valid
+        ) {
+            return Err(Error::Other("authorization not pending or valid".into()));
+        }
+
+        #[derive(Serialize)]
+        struct DeactivateRequest<'a> {
+            status: &'a AuthorizationStatus,
```
djc, line 277: `AuthorizationStatus` is `Copy`, so I think we can just make this owned?

src/lib.rs
```diff
@@ -252,6 +252,52 @@ impl Authorization {
...
+            .await?;
+
+        self.nonce = nonce_from_response(&rsp);
+        self.state = Problem::check::<AuthorizationState>(rsp).await?;
+        if self.state.status != AuthorizationStatus::Deactivated {
+            return Err(Error::Other(
+                "authorization was not deactivated by ACME server".into(),
+            ));
+        }
+        Ok(&self.state)
```
djc, line 298: Nit: would suggest doing this like

```rust
match self.state.status {
     AuthorizationStatus::Deactivated => Ok(&self.state),
     _ => Err(..),
}
```

src/lib.rs
```diff
@@ -70,10 +70,15 @@ impl Order {
     /// After the challenges have been set up, check the [`Order::state()`] to see
     /// if the order is ready to be finalized (or becomes invalid). Once it is
     /// ready, call `Order::finalize()` to get the certificate.
-    pub async fn authorizations(&mut self) -> Result<Vec<AuthorizationState>, Error> {
+    pub async fn authorizations(&mut self) -> Result<Vec<Authorization>, Error> {
```
djc, line 73: I think this is the right direction, but I had something slightly different in mind, I guess. I was wondering if we could have `authorizations()` yield values that are part of the `OrderState`, I guess? Not sure if that would complicate other things with the borrows...

```rust
struct OrderState {
     authorizations: Vec<Authorization>,
     .. 
}

// Name TBD
struct AuthorizationApi<'a> {
     account: &'a AccountInner,
     nonce: &'a mut Option<String>,
     authorization: &'a mut Authorization,
}

impl<'a> Deref for AuthorizationWrap<'a> {
     type Target = Authorization;

     fn deref(&self) -> &Self::Target {
          &*self.authorization
     }
}

impl<'a> DerefMut for AuthorizationWrap<'a> {
     ..
}

struct Authorization {
    pub url: String,
    pub state: Option<AuthorizationState>,
}
```

What do you think? Could also play with this after this lands if you prefer.

djc (commented):
Nice!

src/lib.rs
cpu: >  I was wondering if we could have authorizations() yield values that are part of the OrderState?

I'm not sure I see the advantage. Can you help me understand what properties you're trying to achieve?

Is the idea that the `OrderState` should remain the owner of the authorization state so that we can do more of the processing internal to that type in the future (?)

> What do you think? Could also play with this after this lands if you prefer.

It might be easier for you to just do it vs. catch me up on what you're seeing that I'm not :laughing: but if you're too busy I can probably find time to iterate more this weekend.

src/lib.rs
cpu: Fixed.

src/lib.rs
cpu: Fixed.

src/lib.rs
djc: I guess I don't like the `authorizations()` API much -- it requires allocating a `Vec` with stuff that feels like it should hang off of the `Order` without being directly part of it.

I think this is a good improvement though which I can play with later.

djc (approved):

djc:
Good stuff!

PR #90, opened 2025-03-07
tidy examples, add tls-alpn-01 responder example

cpu:
Opening as a draft for now while there's some questions on my mind:

1. Is this something that makes a worthwhile example to have in-repo?
2. Would it be better to fold into the existing `provision.rs` example?
3. Should we postpone this until we land some more utility helpers in the main library to reduce the duplication between `provision.rs`, `pebble.rs` and this new example?

I mainly implemented this to have a convenient way to test the ARI code with Let's Encrypt staging/production for https://github.com/djc/instant-acme/pull/85 (_though I've backed out the ARI bits to simplify the example_). For that testing I wanted a fully self-contained example that worked with just a DNS name + open port 443. Unlike `provision.rs` it handles the challenge response internally without needing a human operator to do any external setup.

WDYT?

### DIFF
--- a/examples/provision.rs
+++ b/examples/provision.rs
@@ -37,9 +37,13 @@ async fn main() -> anyhow::Result<()> {
     // Create the ACME order based on the given domain names.
     // Note that this only needs an `&Account`, so the library will let you
     // process multiple orders in parallel for a single account.
-    let identifier = Identifier::Dns(opts.name.clone());
+    let identifiers = opts
+        .name
+        .iter()
+        .map(|ident| Identifier::Dns(ident.clone()))
+        .collect::<Vec<_>>();
     let mut order = account
-        .new_order(&NewOrder::new(&[identifier]))
+        .new_order(&NewOrder::new(identifiers.as_slice()))
         .await?;
 
     let state = order.state();
@@ -92,7 +96,7 @@ async fn main() -> anyhow::Result<()> {
 
     // If the order is ready, we can provision the certificate.
     // Use the rcgen library to create a Certificate Signing Request.
-    let mut params = CertificateParams::new(vec![opts.name])?;
+    let mut params = CertificateParams::new(opts.name)?;
     params.distinguished_name = DistinguishedName::new();
     let private_key = KeyPair::generate()?;
     let csr = params.serialize_request(&private_key)?;
@@ -114,5 +118,5 @@ async fn main() -> anyhow::Result<()> {
 #[derive(Parser)]
 struct Options {
     #[clap(long)]
-    name: String,
+    name: Vec<String>,
 }
--- a/examples/provision.rs
+++ b/examples/provision.rs
@@ -18,7 +18,6 @@ async fn main() -> anyhow::Result<()> {
     // Create a new account. This will generate a fresh ECDSA key for you.
     // Alternatively, restore an account from serialized credentials by
     // using `Account::from_credentials()`.
-
     let (account, credentials) = Account::create(
         &NewAccount {
             contact: &[],
@@ -37,7 +36,6 @@ async fn main() -> anyhow::Result<()> {
     // Create the ACME order based on the given domain names.
     // Note that this only needs an `&Account`, so the library will let you
     // process multiple orders in parallel for a single account.
-
     let identifier = Identifier::Dns(opts.name.clone());
     let mut order = account
         .new_order(&NewOrder::new(&[identifier]))
@@ -48,7 +46,6 @@ async fn main() -> anyhow::Result<()> {
     assert!(matches!(state.status, OrderStatus::Pending));
 
     // Pick the desired challenge type and prepare the response.
-
     let authorizations = order.authorizations().await?;
     let mut challenges = Vec::with_capacity(authorizations.len());
     for authz in &authorizations {
@@ -60,7 +57,6 @@ async fn main() -> anyhow::Result<()> {
 
         // We'll use the DNS challenges for this example, but you could
         // pick something else to use here.
-
         let challenge = authz
             .challenges
             .iter()
@@ -83,13 +79,11 @@ async fn main() -> anyhow::Result<()> {
     }
 
     // Let the server know we're ready to accept the challenges.
-
     for url in &challenges {
         order.set_challenge_ready(url).await?;
     }
 
     // Exponentially back off until the order becomes ready or invalid.
-
     let status = order.poll(5, Duration::from_millis(250)).await?;
     if status != OrderStatus::Ready {
         return Err(anyhow::anyhow!("unexpected order status: {status:?}"));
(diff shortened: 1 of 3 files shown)


examples/provision.rs
```diff
@@ -18,7 +18,6 @@ async fn main() -> anyhow::Result<()> {
     // Create a new account. This will generate a fresh ECDSA key for you.
     // Alternatively, restore an account from serialized credentials by
     // using `Account::from_credentials()`.
-
```
djc, line 21: Ahh, I kinda like the whitespace for this.

examples/provision.rs
```diff
@@ -114,5 +118,5 @@ async fn main() -> anyhow::Result<()> {
 #[derive(Parser)]
 struct Options {
     #[clap(long)]
-    name: String,
+    name: Vec<String>,
```
djc, line 121: Nit: let's rename this to `names`?

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+use clap::Parser;
+use hyper_util::client::legacy::Client as HyperClient;
+use hyper_util::rt::TokioExecutor;
+use rcgen::{CertificateParams, CustomExtension, DistinguishedName, KeyPair};
+use rustls::crypto::CryptoProvider;
+use rustls::server::{Acceptor, ClientHello, ResolvesServerCert};
+use rustls::sign::CertifiedKey;
+use rustls::{ClientConfig, RootCertStore, ServerConfig};
+use rustls_pki_types::pem::PemObject;
+use rustls_pki_types::{CertificateDer, PrivatePkcs8KeyDer};
```
djc, line 40: Nit: import `pki_types` via `rustls`?

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+
+        // Complete the handshake using the server config.
+        // There's no application data to care about reading.
+        let _ = handshake.into_stream(config.into()).await;
+    }
+}
+
+/// Simple [`ResolvesServerCert`] implementation that always uses the wrapped CertifiedKey
+///
+/// We can't use `ServerConfig::with_single_cert()` directly because it parses the certificate
```
djc, line 353: Can't we use `SingleCertAndKey::from(certified_key)`?

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+    CertifiedKey::new(vec![challenge_cert_der.clone()], key)
+}
+
+/// A tokio-rustls server that responds to TLS-ALPN-01 challenges
+///
+/// When a client hello with the correct ALPN is received, with a server name indicator (SNI)
+/// extension matching a challenge identifier, the server will respond with a certificate that
+/// contains the SNI as a SAN and the TLS-ALPN-01 challenge response key auth digest extension.
+///
+/// Other types of TLS connection are ignored. In a more complete implementation, the server would
```
djc, line 256: Nit: "connection" -> "connections".

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+};
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::fmt::init();
+
+    let opts = Options::parse();
+
+    // Build a Rustls config that can validate the ACME server's certificate.
+    let rustls_config = if let Some(roots_pem) = &opts.server_root_cert_path {
```
djc, line 61: Why have the `server_root_cert_path` option, is that for local testing? Seems overkill for the example.

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+            .with_tls_config(rustls_config)
+            .https_or_http()
+            .enable_http1()
+            .enable_http2()
+            .build(),
+    );
+
+    // Spawn a challenge response server that listens for incoming TLS connections to solve
+    // TLS-ALPN-01 challenges.
+    let (shutdown_tx, shutdown_rx) = mpsc::channel(1);
```
djc, line 85: Maybe use [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) for this stuff?

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+/// likely want to use a default server configuration for these connections, and process the
+/// transmitted application data in some way.
+struct ChallengeResponseServer {
+    challenges: Arc<Mutex<HashMap<String, Arc<CertifiedKey>>>>,
+    shutdown_rx: mpsc::Receiver<()>,
+}
+
+impl ChallengeResponseServer {
+    async fn run(mut self, port: u16) {
+        info!("starting challenge response server on port: {port}");
```
djc, line 266: Nit: structured logging? (Also, elsewhere.)

examples/tls_alpn_provision.rs
```diff
@@ -0,0 +1,363 @@
...
+                    break;
+                },
+            }
+        }
+
+        info!("shutting down challenge response server");
+    }
+
+    async fn handle_connection(&self, stream: TcpStream, addr: SocketAddr) {
+        info!("handling conn from {}", addr);
```
djc, line 292: Inline format arguments?

djc (commented):
I think it would be good to have something like this. On the other hand, I feel like maybe we should actually build a full-blown server that would be easy to plug, say, Axum things into, which builds in hyper-rustls and integrated ACME and all the niceties instead. I've actually started on a hyper-rustls server but it might end up being a little too opinionated for that project.

It also feels like we should probably do something about all the duplication before we merge this.

examples/provision.rs
cpu: NP. I'll revert this.

examples/tls_alpn_provision.rs
cpu: > Why have the server_root_cert_path option, is that for local testing?

Yeah exactly, it was handy to do the initial development/testing locally against Pebble. It might also be handy to allow running the example in CI (but didn't implement that in this branch).

I don't feel strongly if you prefer it left out.

cpu:
> On the other hand, I feel like maybe we should actually build a full-blown server that would be easy to plug, say, Axum things into, which builds in hyper-rustls and integrated ACME and all the niceties instead. I've actually started on a hyper-rustls server but it might end up being a little too opinionated for that project.

Agreed that it would be better to find a way to plug into an actual server framework. I think the main disadvantage of this example is that it doesn't show how you would do this _and_ have a real application on :443. Getting that working in an Axum application with a dummy health-check endpoint or something would be a more compelling/useful example.

> It also feels like we should probably do something about all the duplication before we merge this.

Sounds good. It feels like what's missing is a way to take an `Order` and a `ChallengeType` and have everything except the challenge response provisioning handled for you. Maybe the `AuthorizationMethod` trait from `pebble.rs` could do the job if we removed the `PATH` parts that are specific to the challenge test server?

> (Happy to take all/most of the cleanup before the last commit!)

:+1: I'll close this for now then and break out just the cleanup commits.

PR #92, opened 2025-03-11
Scope API to wrapper types

djc:
Here's an attempt at making the API a little more restricted while also moving more logic into the library. I quite like it, but curious for your thoughts. It also helps a little with the boilerplate across examples/tests.


cpu:
I'll take a look at this soon. Thanks!

src/lib.rs
```diff
@@ -226,30 +185,72 @@ impl Order {
...
 ///
 /// Authorizations are retrieved from an associated [`Order`] by calling
-/// [`Order::authorizations()`].
+/// [`Order::authorizations()`]. This type dereferences to the underlying
+/// [`AuthorizationState`] for easy access to the authorization's state.
+///
+/// For each authorization, you'll need to:
+///
+/// * Select which [`ChallengeType`] you want to complete
+/// * Call [`AuthorizationHandle::challenge()`] to get a [`ChallengeHandle`]
```
cpu, line 233: maybe missing another bullet point saying something like "* Use the `ChallengeHandle` to complete the authorization's challenge" ?

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+        self.state
+    }
+}
+
+/// A [`Challenge`] wrapper that exposes relevant API
+///
+/// For each challenge, you'll need to:
+///
+/// * Obtain the [`ChallengeHandle::key_authorization()`] for the challenge response
+/// * Set up the challenge response in your infrastructure
```
cpu, line 334: editorial:

```suggestion
/// * Set up the challenge response in your infrastructure (details vary by challenge type)
```

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+/// * Call [`ChallengeHandle::set_ready()`] for that challenge
```
cpu, line 335: editorial: maybe helpful to emphasis this is the point where you're asking the remote peer to do something that can't be undone.

```suggestion
/// * Call [`ChallengeHandle::set_ready()`] for that challenge once you are ready for the ACME server to verify it
```

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+///
+/// After the challenges have been set up, call [`Order::poll()`] to wait until the order is
```
cpu, line 337: Once the challenge is set up, you call set_ready(), _then_ you poll:

```suggestion
/// After the challenges has been set ready, call [`Order::poll()`] to wait until the order is
```

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+/// ready to be finalized (or if it becomes invalid). Once it is ready, call [`Order::finalize()`]
```
cpu, line 338: editorial:

```suggestion
/// ready to be finalized (or to learn if it becomes invalid). Once it is ready, call [`Order::finalize()`]
```

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+    identifier: &'a Identifier,
+    challenge: &'a Challenge,
+    nonce: &'a mut Option<String>,
+    account: &'a AccountInner,
+}
+
+impl ChallengeHandle<'_> {
+    /// Notify the server that the given challenge is ready to be completed
+    ///
+    /// `challenge_url` should be the `Challenge::url` field.
```
cpu, line 352: I think this bit about a `challenge_url` in the docstring is stale now right? No arguments are being passed in by the caller.

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+            .await?;
+
+        *self.nonce = nonce_from_response(&rsp);
+        let _ = Problem::check::<Challenge>(rsp).await?;
+        Ok(())
+    }
+
+    /// Create a [`KeyAuthorization`] for this challenge
+    ///
+    /// Signs the challenge's token with the account's private key and use the
```
cpu, line 366: There's no signature in a key authorization. It's a combination of a challenge token from the ACME server and the encoding of the "thumbprint" digest hash of the account key public key (https://www.rfc-editor.org/rfc/rfc8555#section-8.1).

Maybe we could say something like:

```rust
/// Combines a challenge's token with with the thumbprint of the account's public key to compute the
/// challenge `KeyAuthorization`.
///
/// The derived `KeyAuthorization` can be used to provision the expected challenge response
/// based on the challenge type in use.
```

djc:
Cool! What do you think of the `Handle` suffixes? I don't love them but couldn't think of anything better.

cpu:
> Cool! What do you think of the Handle suffixes? I don't love them but couldn't think of anything better.

I felt the same. Any alternatives I could think up felt even more awkward.

src/lib.rs
```diff
@@ -226,30 +185,72 @@ impl Order {
...
+                Ok(state) => authz.state = Some(state),
+                Err(e) => return Some(Err(e)),
+            }
+        }
+
+        Some(Ok(AuthorizationHandle {
+            // This is a little ugly, but I don't think we can do better in 2021 edition.
+            // Seems like there are other better options with the scope changes in 2024,
+            // which we can't use yet due to MSRV limitations.
+            state: authz.state.as_mut().unwrap(),
```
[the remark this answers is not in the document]
djc: Too lazy to file an issue -- leaving the "MSRV" stanza in to allow a search to trigger this upon a future MSRV bump.

src/lib.rs
```diff
@@ -284,30 +285,101 @@ impl Authorization {
...
+
+impl Deref for AuthorizationHandle<'_> {
+    type Target = AuthorizationState;
+
+    fn deref(&self) -> &Self::Target {
+        self.state
+    }
+}
+
+/// A [`Challenge`] wrapper that exposes relevant API
```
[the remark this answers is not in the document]
djc: Rewritten as `Wrapper type for interacting with a [`Challenge`]'s state`.

src/lib.rs
djc: Added "after setup is complete", would like to keep it a bit shorter.

djc:
Merged this but happy to take additional feedback on the comments/changes.

Issue #93, opened 2025-03-17
Account should be able to be created from a key only


axos88:
According to https://github.com/algesten/acme-lib/blob/5a4d55080635dc9a7f9a380a79984e8b5f9ab281/src/dir.rs#L117C9-L119C61

A call to newAccount is fine to do with an already existing key. The call will return the account url in this case.
This feature would allow users to only persist the account key (and the directory url if needed), and retrieve all necessary info from that.

cpu:
> According to https://github.com/algesten/acme-lib/blob/5a4d55080635dc9a7f9a380a79984e8b5f9ab281/src/dir.rs#L117C9-L119C61

I think a better citation for this is [RFC 8555 7.3](https://www.rfc-editor.org/rfc/rfc8555#section-7.3). In particular to do what you're describing it would be important to set the `onlyReturnExisting` bool to `true` in the new account request.

axos88:
Thanks for that, submitting a PR soon

PR #94, opened 2025-03-17
add feature to load an account by private key

axos88:
Addresses #93

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,7 +389,7 @@ impl Deref for ChallengeHandle<'_> {
 /// Create an [`Account`] with [`Account::create()`] or restore it from serialized data
 /// by passing deserialized [`AccountCredentials`] to [`Account::from_credentials()`].
 ///
-/// Alternatively, you can load an account using the private key using [`Account::load()`].
+/// Alternatively, you can load an account using the private key using [`Account::loadgit ()`].
 ///
 /// The [`Account`] type is cheap to clone.
 ///
@@ -511,7 +511,7 @@ impl Account {
         Self::create_inner(
             &ignored_account, // This field is ignored as per rfc8555 7.3.1
             (key, pkcs8),
-            None,             // This field is ignored as per rfc8555 7.3.1
+            None, // This field is ignored as per rfc8555 7.3.1
             client,
             server_url,
         )
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,7 @@ use serde::de::DeserializeOwned;
 use tokio::time::sleep;
 
 mod types;
+use crate::types::LOAD_EXISTING_ACCOUNT;
 #[cfg(feature = "time")]
 pub use types::RenewalInfo;
 pub use types::{
@@ -425,6 +426,48 @@ impl Account {
         })
     }
 
+    /// Load a new account by private key, with a default or custom HTTP client
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
+    ///
+    /// the `DefaultClient::try_new()?` can be used as the http client by default.
+    /// The returned [`AccountCredentials`] can be serialized and stored for later use.
+    /// Use [`Account::from_credentials()`] to restore the account from the credentials.
+    #[cfg(feature = "hyper-rustls")]
+    pub async fn for_existing_account(
+        key: Key,
+        server_url: &str,
+    ) -> Result<(Account, AccountCredentials), Error> {
+        let http = Box::new(DefaultClient::try_new()?);
+
+        Self::for_existing_account_http(key, server_url, http)
+    }
+
+    /// Load a new account by private key, with a default or custom HTTP client
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
+    ///
+    /// the `DefaultClient::try_new()?` can be used as the http client by default.
+    /// The returned [`AccountCredentials`] can be serialized and stored for later use.
+    /// Use [`Account::from_credentials()`] to restore the account from the credentials.
+    pub async fn for_existing_account_http(
+        key: Key,
+        server_url: &str,
+        http: Box<dyn HttpClient>,
+    ) -> Result<(Account, AccountCredentials), Error> {
+        let client = Client::new(server_url, http).await?;
+        let pkcs8 = key.to_pkcs8_der()?;
+
+        Self::create_inner(
+            &LOAD_EXISTING_ACCOUNT,
+            (key, pkcs8),
+            None,
+            client,
+            server_url,
+        )
+        .await
+    }
+
     /// Restore an existing account from the given ID, private key, server URL and HTTP client
     ///
     /// The key must be provided in DER-encoded PKCS#8. This is usually how ECDSA keys are
@@ -479,40 +522,6 @@ impl Account {
             Key::generate()?,
             external_account,
             Client::new(server_url, http).await?,
-            server_url,
-        )
-        .await
-    }
-
-    /// Load a new account by private key, with a default or custom HTTP client
-    ///
-    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
-    ///
-    /// The returned [`AccountCredentials`] can be serialized and stored for later use.
-    /// Use [`Account::from_credentials()`] to restore the account from the credentials.
-    pub async fn load(
-        private_key_pkcs8_der: &[u8],
-        server_url: &str,
-        http: Option<Box<dyn HttpClient>>,
-    ) -> Result<(Account, AccountCredentials), Error> {
-        let client = match http {
-            Some(http) => Client::new(server_url, http).await?,
-            None => Client::new(server_url, Box::new(DefaultClient::try_new()?)).await?,
-        };
-
-        let key = Key::from_pkcs8_der(private_key_pkcs8_der)?;
-        let pkcs8 = key.to_pkcs8_der()?;
-        let ignored_account = NewAccount {
-            only_return_existing: true,
-            contact: &[],
-            terms_of_service_agreed: true,
-        };
-
-        Self::create_inner(
-            &ignored_account, // This field is ignored as per rfc8555 7.3.1
-            (key, pkcs8),
-            None, // This field is ignored as per rfc8555 7.3.1
-            client,
             server_url,
         )
         .await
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -473,20 +476,55 @@ impl Account {
...
+    /// Load a new account by private key, with a default or custom HTTP client
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
+    ///
+    /// The returned [`AccountCredentials`] can be serialized and stored for later use.
+    /// Use [`Account::from_credentials()`] to restore the account from the credentials.
+    pub async fn load(
+        private_key_pkcs8_der: &[u8],
+        server_url: &str,
+        http: Option<Box<dyn HttpClient>>,
```
djc, line 496: Not sure the `Option` type really makes sense for `http`? If it is `Some` you don't want to pull in the `DefaultClient` dependendencies; if it is `None` you need to. Which one do you want to use?

src/lib.rs
```diff
@@ -473,20 +476,55 @@ impl Account {
...
+    ) -> Result<(Account, AccountCredentials), Error> {
+        let client = match http {
+            Some(http) => Client::new(server_url, http).await?,
+            None => Client::new(server_url, Box::new(DefaultClient::try_new()?)).await?,
+        };
+
+        let key = Key::from_pkcs8_der(private_key_pkcs8_der)?;
```
djc, line 503: I think we should probably externalize this, instead make `Key::from_pkcs8_der()` public, and have this function take a `Key`.

src/lib.rs
```diff
@@ -473,20 +476,55 @@ impl Account {
...
+        let ignored_account = NewAccount {
+            only_return_existing: true,
+            contact: &[],
+            terms_of_service_agreed: true,
+        };
+
+        Self::create_inner(
+            &ignored_account, // This field is ignored as per rfc8555 7.3.1
+            (key, pkcs8),
+            None,             // This field is ignored as per rfc8555 7.3.1
```
djc, line 514: These comments don't quite make sense in their current form.

src/lib.rs
```diff
@@ -473,20 +476,55 @@ impl Account {
...
         .await
     }
 
+    /// Load a new account by private key, with a default or custom HTTP client
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
+    ///
+    /// The returned [`AccountCredentials`] can be serialized and stored for later use.
+    /// Use [`Account::from_credentials()`] to restore the account from the credentials.
+    pub async fn load(
```
djc, line 493: Let's call this `for_existing_account()` and move it up above `from_parts()`?

src/lib.rs
axos88: I first did it that way, but wasn't sure you were open to extend the area of public API.

src/lib.rs
axos88: The first comment is there to explain why a dummy ignored_account is created and why sending that as part of the request is okay. The second one might need to be removed - agreed.

src/lib.rs
axos88: Sure.

src/lib.rs
axos88: But to be honest, if we are making Key part of the public API, it would make more sense to add the Key as the param of the Account::create as well, so that the user has the ability to create an account with a chosen key instead of forcing to generate one for them.

And in that case for_existing_account would probably just be a convenience wrapper around that.

src/lib.rs
axos88: Ahh, I did not see that was behind a feature gate.

axos88:
How can I run the pebble test locally?

src/lib.rs
```diff
@@ -425,6 +426,48 @@ impl Account {
...
     }
 
+    /// Load a new account by private key, with a default or custom HTTP client
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
+    ///
+    /// the `DefaultClient::try_new()?` can be used as the http client by default.
+    /// The returned [`AccountCredentials`] can be serialized and stored for later use.
+    /// Use [`Account::from_credentials()`] to restore the account from the credentials.
+    #[cfg(feature = "hyper-rustls")]
```
djc, line 436: Still don't really want to have both the `_http()` and non-`_http()` variants. Which one of these do you want to use?

src/lib.rs
```diff
@@ -907,22 +916,22 @@ struct Key {
 }
 
 impl Key {
-    fn generate() -> Result<(Self, crypto::pkcs8::Document), Error> {
+    pub fn generate() -> Result<(Self, crypto::pkcs8::Document), Error> {
```
djc, line 921: I'm okay with making `generate()` and `from_pkcs8_der()` public. I don't think `new()` and `to_pkcs8_der()` need to become public.

@cpu thoughts?

src/types.rs
```diff
@@ -507,6 +507,16 @@ pub struct NewAccount<'a> {
...
 
+/// This is a special value for `NewAccount` that is supplied
+/// when loading an account from a private key.
+/// According to rfc8555 7.3.1 this field needs to be supplied,
+/// but MUST be ignored by the ACME server.
+pub(crate) static LOAD_EXISTING_ACCOUNT: NewAccount = NewAccount {
+    only_return_existing: true,
+    contact: &[],
+    terms_of_service_agreed: true,
+};
```
djc, line 518: Rename this to:

```rust
impl NewAccount<'_> {
     pub(crate) const EXISTING: Self = Self {
          ..
     };
}
```

djc (commented):
Please squash all your commits.

cpu:
> How can I run the pebble test locally?

It's a little more tricky than it should be at the moment because we're waiting on an upstream Pebble release. The basic process is:

1. Get `pebble` and `pebble-challtestsrv` binaries
2. Put them in the root of your `instant-acme` checkout
3. Run `cargo test --features=x509-parser --features=time -- --ignored`

For step 1, since we're presently waiting on a fresh release of Pebble you'll have to build it from source (which requires Go) for all the tests to pass. You can see how CI does it here: https://github.com/djc/instant-acme/blob/main/.github/workflows/pebble.yml

Alternatively, you could download pre-built releases here: https://github.com/letsencrypt/pebble/releases/tag/v2.7.0 but you'll see that one of the Pebble tests fails (`account_deactivate`) because it's missing an unreleased Pebble fix.

src/lib.rs
```diff
@@ -29,6 +29,7 @@ use serde::de::DeserializeOwned;
 use tokio::time::sleep;
 
 mod types;
+use crate::types::LOAD_EXISTING_ACCOUNT;
```
cpu, line 32: I think this import should be moved into the existing `use types::{ ... }` group on L40..43

src/lib.rs
```diff
@@ -389,6 +390,8 @@ impl Deref for ChallengeHandle<'_> {
 /// Create an [`Account`] with [`Account::create()`] or restore it from serialized data
 /// by passing deserialized [`AccountCredentials`] to [`Account::from_credentials()`].
 ///
+/// Alternatively, you can load an account using the private key using [`Account::loadgit ()`].
```
cpu, line 393: I think some details shifted here and  `Account::loadgit()` isn't a fn?

src/lib.rs
```diff
@@ -423,6 +426,48 @@ impl Account {
         })
     }
 
+    /// Load a new account by private key, with a default or custom HTTP client
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
```
cpu, line 431:
```suggestion
    /// <https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1>
```

src/lib.rs
```diff
@@ -423,6 +426,48 @@ impl Account {
         })
     }
 
+    /// Load a new account by private key, with a default or custom HTTP client
```
cpu, line 429: I don't think this headline is accurate for this fn: it always uses the default HTTP client.

src/lib.rs
```diff
@@ -423,6 +426,48 @@ impl Account {
...
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc8555#section-7.3.1
+    ///
+    /// the `DefaultClient::try_new()?` can be used as the http client by default.
```
cpu, line 433: similar to my prev comment, for this fn (if both are kept):

```suggestion
    /// the `DefaultClient::try_new()?` will be used as the http client.
```

src/lib.rs
```diff
@@ -861,26 +908,35 @@ impl fmt::Debug for Client {
     }
 }
 
-struct Key {
+/// This struct represents a key used to sign requests to the ACME server
```
cpu, line 911: "Key" can be fairly ambiguous in this context :-) Let's explicitly say it's the **account** private key.

```suggestion
/// This struct represents an account private key used to sign requests to the ACME server
```

src/lib.rs
cpu: > @cpu thoughts?

Yeah, in the absence of some additional motivation I agree :+1:

src/lib.rs
```diff
@@ -861,26 +908,35 @@ impl fmt::Debug for Client {
...
+    /// Generate a random key
+    pub fn generate() -> Result<(Self, crypto::pkcs8::Document), Error> {
         let rng = crypto::SystemRandom::new();
         let pkcs8 =
             crypto::EcdsaKeyPair::generate_pkcs8(&crypto::ECDSA_P256_SHA256_FIXED_SIGNING, &rng)?;
         Self::new(pkcs8.as_ref(), rng).map(|key| (key, pkcs8))
     }
 
-    fn from_pkcs8_der(pkcs8_der: &[u8]) -> Result<Self, Error> {
+    /// Create a key from a der encoded ES256
```
cpu, line 928: I think we can be more specific here:

```suggestion
    /// Create a `Key` from a DER encoded PKCS8 format ECDSA P-256 private key
```

src/lib.rs
```diff
@@ -861,26 +908,35 @@ impl fmt::Debug for Client {
...
+    pub fn from_pkcs8_der(pkcs8_der: &[u8]) -> Result<Self, Error> {
         Self::new(pkcs8_der, crypto::SystemRandom::new())
     }
 
-    fn new(pkcs8_der: &[u8], rng: crypto::SystemRandom) -> Result<Self, Error> {
+    /// Export a key to a der encoded ES256 key
```
cpu, line 933: If this fn stays I think similar feedback as the `from_pkcs8_der()` specificity applies.

src/lib.rs
```diff
@@ -861,26 +908,35 @@ impl fmt::Debug for Client {
...
+    pub fn to_pkcs8_der(&self) -> Result<crypto::pkcs8::Document, Error> {
+        Ok(self.inner.to_pkcs8v1()?)
+    }
+
+    /// Create a key from a der encoded ES256 key and a crypto::SystemRandom
```
cpu, line 938: If this fn stays I think similar feedback as the `from_pkcs8_der()` specificity applies.

src/types.rs
```diff
@@ -507,6 +507,16 @@ pub struct NewAccount<'a> {
     pub only_return_existing: bool,
 }
 
+/// This is a special value for `NewAccount` that is supplied
+/// when loading an account from a private key.
+/// According to rfc8555 7.3.1 this field needs to be supplied,
+/// but MUST be ignored by the ACME server.
```
cpu, line 513: I think we can describe this a bit clearer:

```suggestion
/// A `NewAccount`  endpoint request payload for fetching an existing account.
///
/// `only_return_existing` is set to true to avoid creating a new account if the
/// existing account is not known to the ACME server for some reason.
```

tests/pebble.rs
```diff
@@ -286,6 +286,60 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
     Ok(())
 }
 
+/// Test account loading by private key
+#[tokio::test]
+#[ignore]
+async fn account_from_private_key() -> Result<(), Box<dyn StdError>> {
+    use serde_json::Value;
+    use instant_acme::Key;
```
cpu, line 294: Let's import these top-level

cpu (commented):
Cool, I think this is coming along :+1:

src/lib.rs
axos88: Lol, i guess my cli command got there by mistake

axos88:
Thanks for the feedback I'll reiterate in a few days when i get some free time

georgewhewell:
> Thanks for the feedback I'll reiterate in a few days when i get some free time

any update? @axos88 @cpu 
thanks for the pr already, looking forward to this feature- i can take over if you are unable 👍

cpu:
@georgewhewell Since we haven't heard back from the original author in a ~week it seems fair game to me if you want to start a new branch/PR taking over from where we left off.

axos88:
Crap, sorry guys, was extremely busy this summer. Thanks for getting it done, @djc

PR #95, opened 2025-03-17
Add simplified Order::finalize() method

djc:
Slicing off some more boilerplate. Doesn't feel like a big win, but might be a small one. What do you think?


src/lib.rs
```diff
@@ -69,11 +69,34 @@ impl Order {
         }
     }
 
+    /// Generate a Certificate Signing Request for the given names and request finalization
```
cpu, line 72: One refactor I was hoping to figure out is to have the `Order` know the `Identifier`'s it was created with. After that I think the desired API shape for this operations would be something like `let cert_pem = my_order.finalize().await?` without any args. 

Asking the user to provide the `Identifiers` at the time of finalization like this API will be prone to error. [RFC 8555 7.4](https://www.rfc-editor.org/rfc/rfc8555#section-7.4) says the server has to return an error if the CSR's names and the order's identifiers aren't an exact match.

src/lib.rs
```diff
@@ -69,11 +69,34 @@ impl Order {
...
+    ///
+    /// Uses the rcgen crate to generate a Certificate Signing Request (CSR) for the given
+    /// Subject Alternative Names, then calls [`Order::finalize()`] with it. Returns the generated
+    /// private key, serialized as PEM.
+    ///
+    /// After this succeeds, call [`Order::certificate()`] to retrieve the certificate chain once
+    /// the order is in the appropriate state.
+    #[cfg(feature = "rcgen")]
+    pub async fn finalize_for(&mut self, names: Vec<String>) -> Result<String, Error> {
+        use rcgen::{CertificateParams, DistinguishedName, KeyPair};
```
cpu, line 82: :astonished: I thought you always wanted this sort of import top-level!

src/lib.rs
```diff
@@ -69,11 +69,34 @@ impl Order {
...
+
+        let mut params = CertificateParams::new(names)?;
+        params.distinguished_name = DistinguishedName::new();
```
cpu, line 85: unrelated: I'd love to kill this step upstream.... IMO for EE certs in 2025 not having a common name should be the default.

src/types.rs
```diff
@@ -50,6 +50,10 @@ pub enum Error {
     /// Failed to (de)serialize a JSON object
     #[error("failed to (de)serialize JSON: {0}")]
     Json(#[from] serde_json::Error),
+    /// Failed to build a Certificate Signing Request
+    #[cfg(feature = "rcgen")]
+    #[error("failed to generate CSR: {0}")]
+    Rcgen(#[from] rcgen::Error),
```
cpu, line 56: Will this make updating rcgen across major versions a semver breaking action for this crate?

cpu (commented):
Generally looks good, but I think we can make the API a bit easier. WDYT? I'm also open to landing this as-is and then iterating.

src/lib.rs
djc: Yeah...

src/lib.rs
djc: Fixed this!

src/lib.rs
```diff
@@ -75,12 +77,41 @@ impl Order {
...
+    /// Generate a Certificate Signing Request for the given names and request finalization
+    ///
+    /// Uses the rcgen crate to generate a Certificate Signing Request (CSR) for the given
+    /// Subject Alternative Names, then calls [`Order::finalize()`] with it. Returns the generated
+    /// private key, serialized as PEM.
+    ///
+    /// After this succeeds, call [`Order::certificate()`] to retrieve the certificate chain once
+    /// the order is in the appropriate state.
+    #[cfg(feature = "rcgen")]
+    pub async fn finalize(&mut self) -> Result<String, Error> {
```
cpu, line 89: I wonder if we should collapse this down into one operation that returns both the private key & the certificate in one-go. I'm having a hard time thinking of a reason why you'd want to get just the private key and separately `await` the issued certificate (and the `finalize_csr()` route would let you if needed).

In both `provision.rs` and `pebble.rs` we always immediately fetch the cert after the finalize step so that would reduce a bit more boilerplate IMO.

cpu (approved):
Nice!

src/lib.rs
djc: Makes sense -- but would like to leave that for a future PR.

djc:
> We're collecting up the identifiers from the order authorizations now, but wildcard identifiers are handled specially (ugh) in the authz representation. [RFC 8555 7.1.4](https://www.rfc-editor.org/rfc/rfc8555#section-7.1.4) says:
> 
> > Wildcard domain names (with "*" as the first label) MUST NOT be included in authorization objects.  If an authorization object conveys authorization for the base domain of a newOrder DNS identifier containing a wildcard domain name, then the optional authorizations "wildcard" field MUST be present with a value of true.
> 
> E.g. I think an order that includes `*.example.com` gets an authorization server-side that has the identifier "example.com", with "wildcard: true". When we process the stream for the purpose of making a CSR I think we need to respect that boolean & prepend the wildcard label ourselves.

Oh wow, that's pretty interesting stuff... Reworked the code to try and encode this into the type system, please have a look when you have a moment.

tests/pebble.rs
```diff
@@ -437,13 +437,8 @@ impl Environment {
...
                 .ok_or(format!("no {:?} challenge found", A::TYPE))?;
 
-            let Identifier::Dns(identifier) = challenge.identifier() else {
-                panic!("unsupported identifier type");
-            };
-
             let key_authz = challenge.key_authorization();
-            self.request_challenge::<A>(identifier, &challenge, &key_authz)
-                .await?;
+            self.request_challenge::<A>(&challenge, &key_authz).await?;
```
cpu, line 439: Nice :+1:

src/types.rs
```diff
@@ -568,14 +568,27 @@ fn base64(data: &impl Serialize) -> Result<String, serde_json::Error> {
...
     pub status: AuthorizationStatus,
     /// Possible challenges for the authorization
     pub challenges: Vec<Challenge>,
+    /// Whether the identifier represents a wildcard domain name
+    #[serde(default)]
+    pub wildcard: bool,
+}
+
+impl AuthorizationState {
+    /// Creates a string for the identifier in this authorization
```
cpu, line 593: nit: I think this isn't creating a `String` or `&str`, but a `AuthorizedIdentifier`.

src/types.rs
```diff
@@ -568,14 +568,27 @@ fn base64(data: &impl Serialize) -> Result<String, serde_json::Error> {
...
+    ///
+    /// Takes the `wildcard` field into account.
```
cpu, line 595: I'm not sure this line adds a lot of value. It might be better if it said _how_ it's taken into account, but I think that's already covered by `Identifier::authorized()`.

src/types.rs
```diff
@@ -600,10 +613,34 @@ pub enum Identifier {
...
+    /// wildcard domain.
+    pub fn authorized(&self, wildcard: bool) -> AuthorizedIdentifier<'_> {
+        AuthorizedIdentifier {
+            identifier: self,
+            wildcard,
+        }
+    }
+}
+
+/// An [`Identifier`] which knows it's `wildcard` context
```
cpu, line 636: editorial: I think "its" instead of "it's".

src/lib.rs
```diff
@@ -76,12 +78,41 @@ impl Order {
         }
     }
 
+    /// Generate a Certificate Signing Request for the given names and request finalization
```
cpu, line 81: I think this docstring headline fell out of sync.  Maybe:

```
 /// Generate a Certificate Signing Request for the order's identifiers and request finalization
```

src/lib.rs
```diff
@@ -76,12 +78,41 @@ impl Order {
...
+    ///
+    /// Uses the rcgen crate to generate a Certificate Signing Request (CSR) for the given
+    /// Subject Alternative Names, then calls [`Order::finalize()`] with it. Returns the generated
+    /// private key, serialized as PEM.
```
cpu, line 85: Another bit where we say "given ...". Maybe:

```
 Uses the rcgen crate to generate a Certificate Signing Request (CSR) converting the order's
 identifiers into CSR Subject Alternative Names
```

src/lib.rs
```diff
@@ -76,12 +78,41 @@ impl Order {
...
+    }
+
     /// Request a certificate from the given Certificate Signing Request (CSR)
     ///
-    /// Creating a CSR is outside of the scope of instant-acme. Make sure you pass in a
-    /// DER representation of the CSR in `csr_der`. Call `certificate()` to retrieve the
-    /// certificate chain once the order is in the appropriate state.
-    pub async fn finalize(&mut self, csr_der: &[u8]) -> Result<(), Error> {
+    /// `csr_der` contains the CSR representation serialized in DER encoding. If you don't need
+    /// custom certificate parameters, [`Order::finalize_for()`] can generate the CSR for you.
```
cpu, line 111: ` [`Order::finalize_for()`]` is stale, it's `Order::finalize()` now.

cpu (approved):
Just a few doc nits from the latest updates :+1:

cpu:
> Oh wow, that's pretty interesting stuff...

I wish I remembered why it ended up this way :thinking:

cpu:
:tada: Very cool.

It was also nice to see the Pebble integration tests (and cc92a3e62e9e4be56849574109563252ecc9dec4) pulling their weight and catching a bug before it lands!!

PR #97, opened 2025-03-27
Add RFC 8738 IP address identifiers

cpu:
While [Let's Encrypt has announced upcoming support](https://letsencrypt.org/2025/01/16/6-day-and-ip-certs/#ip-address-support-for-securing-additional-use-cases) for issuing certificates for IP address subjects, it isn't available yet.

Fortunately, Pebble does support this, so we can implement the feature now and have it ready for Let's Encrypt's staging and production environments.

See [RFC 8738](https://www.rfc-editor.org/rfc/rfc8738).

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::fmt;
 use std::fmt::Write;
+use std::net::IpAddr;
 
 use base64::prelude::{BASE64_URL_SAFE_NO_PAD, Engine};
 use bytes::Bytes;
@@ -616,6 +617,11 @@ pub enum AuthorizationStatus {
 #[serde(tag = "type", content = "value", rename_all = "camelCase")]
 pub enum Identifier {
     Dns(String),
+
+    /// An IP address (IPv4 or IPv6) identifier
+    ///
+    /// Note that not all ACME servers will accept an order with an IP address identifier.
+    Ip(IpAddr),
 }
 
 impl Identifier {
@@ -638,6 +644,9 @@ pub struct AuthorizedIdentifier<'a> {
     /// The source identifier, missing any wildcard context
     pub identifier: &'a Identifier,
     /// Whether the identifier should be interpreted as a wildcard
+    ///
+    /// This is only relevant for DNS identifiers and must be false for other
+    /// types of identifiers (e.g. IP addresses).
     pub wildcard: bool,
 }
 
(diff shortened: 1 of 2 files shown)


cpu:
One random thought: should we proactively try and enforce not using DNS-01 for IP identifiers somewhere or is it sufficient to let the CA return a problem if it's attempted? RFC 8738 forbids it, and LE only plans to offer support for HTTP-01 and TLS-ALPN-01.

src/types.rs
```diff
@@ -638,6 +643,9 @@ pub struct AuthorizedIdentifier<'a> {
     /// The source identifier, missing any wildcard context
     pub identifier: &'a Identifier,
     /// Whether the identifier should be interpreted as a wildcard
+    ///
+    /// This is only relevant for DNS identifiers and must be false for other
+    /// types of identifiers (e.g. IP addresses).
```
cpu, line 648: This comes from a combination of [RFC 8555 7.1.4](https://www.rfc-editor.org/rfc/rfc8555#section-7.1.4) where it says (emphasis mine):

> This field MUST be present and true for authorizations created as a result of a newOrder request       **containing a DNS identifier** with a value that was a wildcard domain name.  For other authorizations, it MUST be absent.

and [a verified errata](https://www.rfc-editor.org/errata/eid6364) that further clarified it could be omitted, or explicitly set `false`.

src/types.rs
```diff
@@ -616,6 +616,11 @@ pub enum AuthorizationStatus {
 #[serde(tag = "type", content = "value", rename_all = "camelCase")]
 pub enum Identifier {
     Dns(String),
+
+    /// An IP address (IPv4 or IPv6) identifier
+    ///
+    /// Note that not all ACME servers will accept an order with an IP address identifier.
+    Ip(String),
```
djc, line 623: How about we just store an IpAddr in there?

djc:
> One random thought: should we proactively try and enforce not using DNS-01 for IP identifiers somewhere or is it sufficient to let the CA return a problem if it's attempted? RFC 8738 forbids it, and LE only plans to offer support for HTTP-01 and TLS-ALPN-01.

Typically the server decides what types of challenge are available for a given authorization, right? So I don't think there's anything for us to do here...

src/types.rs

src/types.rs
```diff
@@ -646,6 +655,7 @@ impl fmt::Display for AuthorizedIdentifier<'_> {
         match (self.wildcard, self.identifier) {
             (true, Identifier::Dns(dns)) => f.write_fmt(format_args!("*.{dns}")),
             (false, Identifier::Dns(dns)) => f.write_str(dns),
+            (_, Identifier::Ip(addr)) => f.write_str(&addr.to_string()),
```
djc, line 658: Display::fmt(&addr, f)? write!(f, "{addr}")? f.write_fmt(..)?

cpu: ok

djc (approved):

PR #98, opened 2025-03-28
Profiles

djc:


tests/pebble.rs
```diff
@@ -238,6 +238,20 @@ async fn replacement_order() -> Result<(), Box<dyn StdError>> {
...
+#[ignore]
+async fn profiles() -> Result<(), Box<dyn StdError>> {
+    try_tracing_init();
+
+    // Creat an env/initial account
+    let mut env = Environment::new(EnvironmentConfig::default()).await?;
+    let identifiers = dns_identifiers(&["example.com"]);
+    env.test::<Http01>(&NewOrder::new(&identifiers).profile("shortlived"))
+        .await
+        .map(|_| ())
```
cpu, line 252: WDYT about trying to verify we actually used the right profile? Something like this:

```diff
diff --git a/tests/pebble.rs b/tests/pebble.rs
index 3dc1f07..8b7d063 100644
--- a/tests/pebble.rs
+++ b/tests/pebble.rs
@@ -239,17 +239,22 @@ async fn replacement_order() -> Result<(), Box<dyn StdError>> {
 }

 /// Test order profiles
+#[cfg(all(feature = "x509-parser", feature = "time"))]
 #[tokio::test]
 #[ignore]
 async fn profiles() -> Result<(), Box<dyn StdError>> {
     try_tracing_init();

-    // Creat an env/initial account
     let mut env = Environment::new(EnvironmentConfig::default()).await?;
     let identifiers = dns_identifiers(&["example.com"]);
-    env.test::<Http01>(&NewOrder::new(&identifiers).profile("shortlived"))
-        .await
-        .map(|_| ())
+    let cert = env.test::<Http01>(&NewOrder::new(&identifiers).profile("shortlived"))
+        .await?;
+
+    let (_, cert) = x509_parser::parse_x509_certificate(cert.as_ref())?;
+    let validity = cert.validity.time_to_expiration().unwrap();
+    let default_profile = env.config.pebble.profiles.get("default").unwrap();
+    assert!(validity < default_profile.validity_period);
+    Ok(())
 }

 /// Test that it is possible to deactivate an order's authorizations
 ```

File: src/lib.rs [hunk 2]
```diff
@@ -771,6 +771,23 @@ impl Account {
         Ok(())
     }
 
+    /// Yield the profiles supported according to the account's server directory
+    ///
+    /// The first element of the tuple is the profile name, the second is a URL pointing to a
+    /// human-readable description of the profile.
```
cpu, line 777: I think we can't promise that the 2nd tuple value is a URL, the draft spec says:

> The contents of these human-readable descriptions are up to the CA; for example, they might be prose descriptions of the properties of the profile, or the might be URLs pointing at a documentation site. ACME Clients SHOULD present these profile names and descriptions to their operator during initial setup and at appropriate times thereafter.

src/lib.rs
```diff
@@ -771,6 +771,23 @@ impl Account {
...
+    pub fn profiles(&self) -> impl Iterator<Item = ProfileMeta<'_>> {
+        self.inner
+            .client
+            .directory
+            .meta
+            .profiles
+            .iter()
+            .map(|(name, description)| ProfileMeta {
+                name: &name,
+                description: &description,
```
cpu, line 787: clippy says these can lose the `&` and be simplified to:

```rust
            .map(|(name, description)| ProfileMeta {
                name,
                description,
            })
```

src/types.rs
```diff
@@ -418,6 +419,7 @@ impl<'a> NewOrder<'a> {
         Self {
             identifiers,
             replaces: None,
+            profile: None,
```
cpu, line 422: I think similar to ARI we should also gain a `profile: Option<String>` field on the `OrderState`.

That would let `NewOrder`'s that _don't_ specify a `profile` learn what "default" profile the server assigned their order:

>  If the server is advertizing profiles and receives a newOrder request which does not identify a specific profile, it is RECOMMENDED that the server select a profile and associate it with the new Order object.

Relatedly, the profiles spec doesn't say the server MUST reflect the matching `profile` for new order requests that _do_ specify a profile. That's probably good feedback to give to the author. I think ideally we would enforce it matches like for ARI.

cpu (commented):
Looking good. I had a couple pieces of small feedback to consider.

cpu:
Also maybe worth a README mention as a supported feature? :-)

src/types.rs
djc: The spec isn't specific at all about whether the order state will actually reflect back the selected profile, I think?

cpu: Yeah, I think there's language missing to make that crystal clear. I'm partly inferring it based on the [registry update](https://www.ietf.org/archive/id/draft-aaron-acme-profiles-00.html#section-6.2) and the accompanying Pebble logic:

* Here's where it [assigns a default if one isn't provided](https://github.com/letsencrypt/pebble/blob/dc9bf9d200b49f5b587258ec53948f17417a7579/wfe/wfe.go#L1728-L1744)
* And here's where the profile name (provided _or_ the default choice) is [populated in the order state](https://github.com/letsencrypt/pebble/blob/dc9bf9d200b49f5b587258ec53948f17417a7579/wfe/wfe.go#L1776)

cpu: It's also similar [in the Boulder order JSON](https://github.com/letsencrypt/boulder/blob/53c35ac669b212ebcda95c7956d5a1fb2f0d4b1e/wfe2/wfe.go#L1959)

djc: Also, I couldn't think of a great way to test this.

cpu (approved):

src/types.rs
cpu: Yeah I think the Pebble test code makes it annoying to get at the order state while also using all the shared logic for completing the order. Seems fine to not test it explicitly for now. I think the spec should be updated to be clearer and then we can check it in the `lib.rs` code.

cpu: > The spec isn't specific at all about whether the order state will actually reflect back the selected profile, I think?

@djc It looks like this [got addressed](https://author-tools.ietf.org/iddiff?url2=draft-aaron-acme-profiles-01) in -01 :tada: 

diff

PR #100, opened 2025-04-08
Acme account features

christianhoelzl:
This pull request adds account features and one fix to the instant-acme client:

- A new feature is added to update the contact information of an existing acme account
- A new feature is added to perform an account key rollover of an existing acme account

The contact information that is part of the acme account is according to RFC 8555, section 7.3.2. The contact update payload includes a the contact information as for example an email address.

The account key rollover is according to RFC 8555, section 7.3.5. New credentials are created by the instant-acme client and updated on the acme server. The new credentials are returned to the caller for further usage.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -447,9 +447,7 @@ impl ChallengeHandle<'_> {
 
         *self.nonce = nonce_from_response(&rsp);
 
-        let response = Problem::check::<Challenge>(rsp).await?;
-
-        Problem::check_challenge(response)?;
+        let _ = Problem::check::<Challenge>(rsp).await?;
 
         Ok(())
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -447,9 +447,9 @@ impl ChallengeHandle<'_> {
 
         *self.nonce = nonce_from_response(&rsp);
 
-        let resoponse_payload = Problem::check::<Challenge>(rsp).await?;
+        let response_payload = Problem::check::<Challenge>(rsp).await?;
 
-        if let Some(problem_details) = resoponse_payload.error {
+        if let Some(problem_details) = response_payload.error {
             Err(Error::Api(problem_details))
         } else {
             Ok(())
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -449,10 +449,9 @@ impl ChallengeHandle<'_> {
 
         let response = Problem::check::<Challenge>(rsp).await?;
 
-        match response.error {
-            Some(details) => Err(Error::Api(details)),
-            None => Ok(()),
-        }
+        Problem::check_challenge(response)?;
+
+        Ok(())
     }
 
     /// Create a [`KeyAuthorization`] for this challenge
@@ -794,11 +793,11 @@ impl Account {
 
     /// Account key rollover
     ///
-    /// This is useful if you want to change the acme client key of an existing account, e.g.
-    /// to mitigate the risk of a key compromise. This method crates a new client key and changes
+    /// This is useful if you want to change the ACME account key of an existing account, e.g.
+    /// to mitigate the risk of a key compromise. This method creates a new client key and changes
     /// the key associated with the existing account. In case the key rollover succeeds the new
     /// account credentials are returned for further usage. After that a new Account object with
-    /// the updated client key needs to be crated for further interaction with the acme account.
+    /// the updated client key needs to be crated for further interaction with the ACME account.
     ///
     /// See <https://datatracker.ietf.org/doc/html/rfc8555#section-7.3.5> for more information.
     pub async fn change_key(&self, server_url: &str) -> Result<AccountCredentials, Error> {
@@ -848,7 +847,7 @@ impl Account {
     /// Updates the account contacts
     ///
     /// This is useful if you want to update the contact information of an existing account
-    /// on the acme server. The contacts argument replaces existing contacts on
+    /// on the ACME server. The contacts argument replaces existing contacts on
     /// the server. By providing an empty array the contacts are removed from the server.
     ///
     /// See <https://datatracker.ietf.org/doc/html/rfc8555#section-7.3.2> for more information.
@@ -873,7 +872,7 @@ impl Account {
         let response = Problem::check::<Account>(rsp).await?;
         match response.status {
             AuthorizationStatus::Valid => Ok(()),
-            _ => Err("Unexpected account status after update of account contacts".into()),
+            _ => Err("Unexpected account status after updating contact information".into()),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -823,14 +823,14 @@ impl Account {
             old_key: jwk_old,
         };
 
-        let mut inner_header = new_key.header(Some("nonce"), &new_key_url);
+        let mut inner_header = new_key.header(Some("nonce"), new_key_url);
         inner_header.nonce = None;
 
         let inner_body = JoseJson::new(Some(&payload_inner), inner_header, &new_key)?;
 
         let rsp = self
             .inner
-            .post(Some(&inner_body), None, &new_key_url)
+            .post(Some(&inner_body), None, new_key_url)
             .await?;
 
         let _ = Problem::from_response(rsp).await?;
@@ -873,7 +873,7 @@ impl Account {
         let response = Problem::check::<Account>(rsp).await?;
         match response.status {
             AuthorizationStatus::Valid => Ok(()),
-            _ => Err("Unexpected account status after account key rollover".into()),
+            _ => Err("Unexpected account status after update contacts information".into()),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,8 +40,8 @@ pub use types::{
     RevocationReason, RevocationRequest, ZeroSsl,
 };
 use types::{
-    Directory, Empty, FinalizeRequest, Header, JoseJson, Jwk, KeyOrKeyId, NewAccountPayload,
-    Signer, SigningAlgorithm,
+    AccountContactUpdatePayload, AccountPayload, Directory, Empty, FinalizeRequest, Header,
+    JoseJson, Jwk, KeyOrKeyId, NewAccountPayload, NewKeyPayload, Signer, SigningAlgorithm,
 };
 
 /// An ACME order as described in RFC 8555 (section 7.1.3)
@@ -446,8 +446,14 @@ impl ChallengeHandle<'_> {
             .await?;
 
         *self.nonce = nonce_from_response(&rsp);
-        let _ = Problem::check::<Challenge>(rsp).await?;
-        Ok(())
+
+        let resoponse_payload = Problem::check::<Challenge>(rsp).await?;
+
+        if let Some(problem_details) = resoponse_payload.error {
+            Err(Error::Api(problem_details))
+        } else {
+            Ok(())
+        }
     }
 
     /// Create a [`KeyAuthorization`] for this challenge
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -786,6 +792,58 @@ impl Account {
...
+            key_pkcs8: new_key_pkcs8.as_ref().to_vec(),
+            directory: Some(server_url.to_owned()),
+            urls: None,
+        };
+
+        Ok(credentials)
+    }
+
+    /// Updates the account settings
+    pub async fn update_contact<'a>(&self, contact: &'a [&'a str]) -> Result<(), Error> {
```
djc, line 835: Let's rename this to `update_contacts()`, and rename the argument to `contacts`. Can you add some additional detail in a second paragraph in the docstring to clarify whether this extends or replaces the existing contacts?

src/lib.rs
```diff
@@ -786,6 +792,58 @@ impl Account {
...
+        let payload = AccountContactUpdatePayload { contact };
```
djc, line 836: Let's inline the type definition for `AccountContactUpdatePayload` here, and just call it `Contacts`.

src/lib.rs
```diff
@@ -786,6 +792,58 @@ impl Account {
     pub fn id(&self) -> &str {
         &self.inner.id
     }
+
+    /// Account key rollover
+    pub async fn key_change(&self, server_url: &str) -> Result<AccountCredentials, Error> {
```
djc, line 797: Let's call this `change_key()`.

src/lib.rs
```diff
@@ -786,6 +792,58 @@ impl Account {
...
+        let new_key_url = match self.inner.client.directory.key_change.as_deref() {
+            Some(url) => url,
+            None => return Err("Account key rollover not supported by ACME CA".into()),
+        };
+
+        let (new_key, new_key_pkcs8) = Key::generate()?;
+
+        let jwk_old = Jwk::new(&self.inner.key.inner);
+
+        let payload_inner = NewKeyPayload {
```
djc, line 807: Let's inline the `NewKeyPayload` type definition here, and drop the `Payload` suffix. It looks like `account` could be a `&str` instead of `String`?

src/lib.rs
```diff
@@ -446,8 +446,14 @@ impl ChallengeHandle<'_> {
...
-        let _ = Problem::check::<Challenge>(rsp).await?;
-        Ok(())
+
+        let response_payload = Problem::check::<Challenge>(rsp).await?;
+
+        if let Some(problem_details) = response_payload.error {
+            Err(Error::Api(problem_details))
+        } else {
+            Ok(())
+        }
```
djc, line 456: Nit, drop the `_payload` suffix, the `problem_` prefix and rewrite it like this:

```rust
match payload.error {
    Some(details) => Err(Error::Api(details)),
    None => Ok(()),
}
```

djc (commented):
Thanks for working on this! I think we'll want some tests for this in the `pebble` integration test. Please squash any fixup commits.

Would be helpful to add links to the specific RFC sections in the docstring/comments.

src/lib.rs
christianhoelzl: The type is inline and renamed to Contacts.

src/lib.rs
```diff
@@ -786,6 +792,58 @@ impl Account {
...
+    /// Updates the account settings
+    pub async fn update_contact<'a>(&self, contact: &'a [&'a str]) -> Result<(), Error> {
+        let payload = AccountContactUpdatePayload { contact };
+
+        let rsp = self
+            .inner
+            .post(Some(&payload), None, &self.inner.id)
+            .await?;
+
+        let _ = Problem::check::<AccountPayload>(rsp).await?;
```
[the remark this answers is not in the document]
christianhoelzl: The type definition is inline now. The status is changed to an existing enum AuthorizationStatus and checked. Only Valid is accepted.

src/lib.rs
christianhoelzl: Method name changed.

src/lib.rs
christianhoelzl: Changed, after inlining there are no changes to types.rs anymore.

src/lib.rs
christianhoelzl: changed as proposed

djc:
Just add new tests below the list existing test functions. Suggest just trying out whether Pebble supports these endpoints.

christianhoelzl:
Two pebble tests are added to this pull request. I tested locally on windows OS with the latest pebble release.

src/lib.rs
```diff
@@ -446,8 +446,13 @@ impl ChallengeHandle<'_> {
...
         *self.nonce = nonce_from_response(&rsp);
-        let _ = Problem::check::<Challenge>(rsp).await?;
-        Ok(())
+
+        let response = Problem::check::<Challenge>(rsp).await?;
+
+        match response.error {
+            Some(details) => Err(Error::Api(details)),
+            None => Ok(()),
+        }
```
djc, line 455: This change seems unrelated and introduces consistency issues. If you think this is an improvement, let's do this inside `Problem::from_response()`?

tests/pebble.rs
```diff
@@ -313,6 +313,49 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
...
+#[ignore]
+async fn update_contacts() -> Result<(), Box<dyn StdError>> {
+    try_tracing_init();
+
+    // Creat an env/initial account
+    let env = Environment::new(EnvironmentConfig::default()).await?;
+
+    // Provide empty contacts information, this is fine for pebble
+    let result = env.account.update_contacts(&[]).await?;
+    assert_eq!(result, ());
```
djc, line 326: Nit: this `assert_eq!()` is pointless.

tests/pebble.rs
```diff
@@ -313,6 +313,49 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
...
+
+    // At the time of writing this testcase, pebble does not support contacts information.
+    let Err(Error::Api(problem)) = env.account.update_contacts(&["alice@example.com"]).await else {
+        panic!("unexpected result");
+    };
+
+    assert_eq!(
+        problem.r#type,
```
djc, line 334: Suggest doing `.as_ref()` on `problem.r#type` instead of `to_string()` on the literal.

tests/pebble.rs
```diff
@@ -313,6 +313,49 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
...
+async fn change_key() -> Result<(), Box<dyn StdError>> {
+    try_tracing_init();
+
+    // Creat an env/initial account
+    let mut env = Environment::new(EnvironmentConfig::default()).await?;
+
+    // Change the account key
+    let result = env
+        .account
+        .clone()
```
djc, line 351: Nit: no need to `clone()` here, I think?

tests/pebble.rs
```diff
@@ -313,6 +313,49 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
...
+        .change_key("https://[::1]:14000/dir")
+        .await;
+    assert!(result.is_ok());
```
djc, line 354: Instead of this `assert!()`, just tack a `?` onto the `.await` above.

djc (commented):
Thanks! A few more small issues.

cpu (commented):
Cool :-)

tests/pebble.rs
christianhoelzl: This pointless assert is removed.

tests/pebble.rs
christianhoelzl: done

tests/pebble.rs
christianhoelzl: The clone() is removed

tests/pebble.rs
christianhoelzl: done

src/lib.rs
```diff
@@ -786,6 +791,91 @@ impl Account {
     pub fn id(&self) -> &str {
         &self.inner.id
     }
+
+    /// Account key rollover
+    ///
+    /// This is useful if you want to change the acme client key of an existing account, e.g.
```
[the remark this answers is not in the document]
christianhoelzl: Changed to uppercase ACME and account key.

src/lib.rs
```diff
@@ -786,6 +791,91 @@ impl Account {
...
+    /// to mitigate the risk of a key compromise. This method crates a new client key and changes
+    /// the key associated with the existing account. In case the key rollover succeeds the new
+    /// account credentials are returned for further usage. After that a new Account object with
+    /// the updated client key needs to be crated for further interaction with the acme account.
```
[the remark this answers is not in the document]
christianhoelzl: fixed

src/lib.rs
```diff
@@ -786,6 +791,91 @@ impl Account {
...
+            urls: None,
+        };
+
+        Ok(credentials)
+    }
+
+    /// Updates the account contacts
+    ///
+    /// This is useful if you want to update the contact information of an existing account
+    /// on the acme server. The contacts argument replaces existing contacts on
```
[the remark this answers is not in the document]
christianhoelzl: fixed

src/lib.rs
```diff
@@ -786,6 +791,91 @@ impl Account {
...
+
+        #[derive(Clone, Debug, serde::Deserialize)]
+        struct Account {
+            status: AuthorizationStatus,
+        }
+
+        let response = Problem::check::<Account>(rsp).await?;
+        match response.status {
+            AuthorizationStatus::Valid => Ok(()),
+            _ => Err("Unexpected account status after update contacts information".into()),
```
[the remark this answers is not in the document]
christianhoelzl: I changed it, the new sentence better.

tests/pebble.rs
```diff
@@ -313,6 +313,49 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
...
+    assert_eq!(
+        problem.r#type,
+        Some("urn:ietf:params:acme:error:unsupportedContact".to_string())
+    );
+    Ok(())
+}
+
+#[tokio::test]
+#[ignore]
+async fn change_key() -> Result<(), Box<dyn StdError>> {
```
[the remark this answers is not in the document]
christianhoelzl: The test method is extended to have both cases tested. The first one produces a "malformed" error and the second should succeed.

tests/pebble.rs
```diff
@@ -313,6 +313,49 @@ async fn account_deactivate() -> Result<(), Box<dyn StdError>> {
...
+    try_tracing_init();
+
+    // Creat an env/initial account
+    let env = Environment::new(EnvironmentConfig::default()).await?;
+
+    // Provide empty contacts information, this is fine for pebble
+    let result = env.account.update_contacts(&[]).await?;
+    assert_eq!(result, ());
+
+    // At the time of writing this testcase, pebble does not support contacts information.
```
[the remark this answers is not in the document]
cpu: Ahhh!!! I think that's because you're sending the contact `alice@example.com`. I believe you want `mailto://alice@example.com`. The contacts are **URLs** and so need a scheme, like `mailto://` or `tel://`.

See RFC 8555 7.3, servers MUST support `mailto://` at a minimum:
>  If the server validates contact URLs, it MUST support the "mailto" scheme.

src/lib.rs
christianhoelzl: This seems unrelated. On the other hand RFC8555 defines an inconsistent error handling (chapter 6.7 and 8).

1. Errors reported at the HTTP layer (this is `Problem::from_response`)
2. More specific within the Challenge Object

Adding both to `Problem::from_response` in a strict mode would require to differentiate between the Challenge object and any other response types. In the last commit I added an additional method to Problem impl for case 2 instead and used the new `Problem::check_challenge` inside `set_ready`. WDYT?

christianhoelzl:
Most review changes are fixed in this commit. I extended the readme to show the new features.
Still open is the discussion on Problem::from_response and the discussion on the pebble test for update contacts.

cpu:
> Still open is the discussion on Problem::from_response and the discussion on the pebble test for update contacts.

I left [a comment](https://github.com/djc/instant-acme/pull/100#discussion_r2041120332) about the issue for the pebble test for contacts update. I think you just need to fix the contact you're using in the update.

tests/pebble.rs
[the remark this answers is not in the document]
cpu: Just updated my comment above; the correct form is `mailto:alice@example.com`, not `mailto://alice@example.com`. Sorry about that :-)

Making that change the request succeeds.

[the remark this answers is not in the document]
christianhoelzl: Great, good catch! I fixed it in the pull request.

src/lib.rs
```diff
@@ -446,7 +446,11 @@ impl ChallengeHandle<'_> {
             .await?;
 
         *self.nonce = nonce_from_response(&rsp);
-        let _ = Problem::check::<Challenge>(rsp).await?;
+
+        let response = Problem::check::<Challenge>(rsp).await?;
```
djc, line 450: I still don't understand the motivation for this change. It also seems mostly unrelated to the two account management features you've added here, so I'd suggest you revert these changes (and the method addition supporting it) for this PR and submit a new PR for them with a more detailed motivation for why this is necessary/helps.

djc (approved):

cpu (approved):
Thanks for working through all the review feedback. Looks great.

src/lib.rs
christianhoelzl: Fine, I reverted the set ready change and create a second pull request for this.

cpu:
@djc Thinking we should squash merge this one, oui? Do you want to do the honours?

djc:
Fait accompli!

djc:
Thanks for your contribution! Hopefully this left you feeling motivated to contribute more in the future!

PR #101, opened 2025-04-15
probem details from challenge object

christianhoelzl:
This pull requests improves api error reporting for challenge set_ready.

If an ACME server provides error information on the HTTP layer as a RFC 8707 probem details document, this error information is provided to the caller of the set_ready method as Error::Api(...). But if an ACME server provides the error information as part of the Challenge object (error member, RFC8555 section 8) the current implementation does not forward this information to the caller of set_ready as an Err(Error::Api(...)). This is changed in this pull request.


djc (approved):
Okay, I guess that makes sense!

PR #102, opened 2025-04-17
Yield error state from Order::poll() if set

djc:
Inspired by #101. I don't think there are any other places where it makes sense to yield from the `Order` or `Challenge` `error` fields.


cpu (approved):
Good idea :+1:

PR #103, opened 2025-04-23
Tweaks from account features

djc:
With a fair bit of refactoring to make the code a little easier to navigate.

Followup from:

- #100


cpu (approved):
Nice work :+1:

PR #105, opened 2025-05-14
fix bad nonce

christianhoelzl:
Any ACME CA that generate bad nonce responses often (e.g. Pebble CA 5% of all requests) can lead to unexpected errors in the public API (`"urn:ietf:params:acme:error:badNonce"` Api errors). The current value of retries in `Client::post` gives this  Api error every 0.25% in every call to `Client::post` as part of Pebble testing.

This pull request increases the number of retries in `Client::post` from 3 to 10.
- In case 10 retries are reached, this indicates a bad ACME CA.
- To prevent an infinite loop an error is returned after 10 retries.
- There is no different behaviour for well behaved ACME CA with lower bad nonce probability 
- The Pebble test is stable

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,10 @@ impl Client {
         signer: &impl Signer,
         url: &str,
     ) -> Result<BytesResponse, Error> {
-        let mut retries = 3;
+        // Less than 10 retries without a valid body should yield a valid request body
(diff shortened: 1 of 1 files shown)


djc:
Sorry for the long delay in reviewing...

I don't think this is the right direction. This change affects both production and testing use, and I don't think this is a good change for production uses. Also, we should not change the value and the structure in the same commit.

PR #108, opened 2025-06-22
Encapsulate retry policy in a type

djc:
This type can be constructed and updated in a const context and is mostly opaque so we can evolve what the exact policy looks like without API changes.

### DIFF
--- a/src/order.rs
+++ b/src/order.rs
@@ -1,4 +1,4 @@
-use std::ops::Deref;
+use std::ops::{ControlFlow, Deref};
 use std::sync::Arc;
 use std::time::Duration;
 use std::{fmt, slice};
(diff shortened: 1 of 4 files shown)


src/order.rs
```diff
@@ -491,3 +485,71 @@ impl fmt::Debug for KeyAuthorization {
...
+///
+/// The `Default` implementation will retry 5 times with an initial delay of 250ms.
+#[derive(Debug, Clone, Copy)]
+pub struct RetryPolicy {
+    tries: u8,
+    delay: Duration,
+}
+
+impl RetryPolicy {
+    /// A `const` constructor for the default `RetryPolicy`
```
cpu, line 502: very nit-picky nit: I don't think it adds much to mention the `const` aspect since its self-evident from the fn signature.

cpu (approved):
Nice! Very tidy :stars:

src/order.rs
djc: Fair.

PR #109, opened 2025-06-22
Add support for Retry-After headers in poll()

djc:
As discussed in #104, Pebble doesn't support this so it's untested for now.

cc @christianhoelzl

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ aws-lc-rs = { version = "1.8.0", optional = true }
 base64 = "0.22"
 bytes = "1"
 http = "1"
+httpdate = "1.0.3"
 http-body = "1"
 http-body-util = "0.1.2"
 hyper = { version = "1.3.1", features = ["client", "http1", "http2"], optional = true }
--- a/src/order.rs
+++ b/src/order.rs
@@ -1,9 +1,12 @@
 use std::ops::{ControlFlow, Deref};
+use std::str::FromStr;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 use std::{fmt, slice};
 
 use base64::prelude::{BASE64_URL_SAFE_NO_PAD, Engine};
+use http::header::RETRY_AFTER;
+use httpdate::HttpDate;
 #[cfg(feature = "rcgen")]
 use rcgen::{CertificateParams, DistinguishedName, KeyPair};
 use serde::Serialize;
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -25,6 +25,7 @@ aws-lc-rs = { version = "1.8.0", optional = true }
 base64 = "0.22"
 bytes = "1"
 http = "1"
+httpdate = "1.0.3"
```
cpu, line 28: Is there a reason to use `1.0.3` specifically vs `1`?

src/order.rs
```diff
@@ -542,14 +548,29 @@ struct RetryState {
...
 impl RetryState {
-    async fn wait(&mut self) -> ControlFlow<(), ()> {
+    async fn wait(&mut self, after: Option<SystemTime>) -> ControlFlow<(), ()> {
         if self.tries == 0 {
             return ControlFlow::Break(());
         }
 
+        if let Some(after) = after {
+            let now = SystemTime::now();
+            if let Ok(delay) = after.duration_since(now) {
```
cpu, line 558: nit: inline `now`?

src/order.rs
```diff
@@ -542,14 +548,29 @@ struct RetryState {
...
+        }
+
         sleep(self.delay).await;
         self.delay *= 2;
         self.tries -= 1;
         ControlFlow::Continue(())
     }
 }
+
+fn retry_after(rsp: &BytesResponse) -> Option<SystemTime> {
```
cpu, line 572: I think the `?`/`ok()` dance needed to ultimately extract an `Option<SystemTime>` is good justification for pulling this out into a `fn`, but since it only has one call-site would it be less indirection to inline it into `refresh` like:

```rust
    pub async fn refresh(&mut self) -> Result<&OrderState, Error> {
        fn retry_after(rsp: &BytesResponse) -> Option<SystemTime> {
            Some(SystemTime::from(
                HttpDate::from_str(rsp.parts.headers.get(RETRY_AFTER)?.to_str().ok()?).ok()?,
            ))
        }
        
        let rsp = self
            .account
            .post(None::<&Empty>, self.nonce.take(), &self.url)
            .await?;

        self.nonce = nonce_from_response(&rsp);
        self.retry_after = retry_after(&rsp);
        self.state = Problem::check::<OrderState>(rsp).await?;
        Ok(&self.state)
    }
```

or do you think there will be more call-sites in the future?

cpu (approved):

src/order.rs
```diff
@@ -542,14 +548,29 @@ struct RetryState {
...
+    Some(SystemTime::from(
+        HttpDate::from_str(rsp.parts.headers.get(RETRY_AFTER)?.to_str().ok()?).ok()?,
```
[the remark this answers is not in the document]
djc: Thanks, added support for the `delay-seconds` case.

PR #115, opened 2025-06-26
backpressure factor

christianhoelzl:
This pull request makes the backpressure factor configurable.


src/order.rs
```diff
@@ -496,6 +496,7 @@ impl fmt::Debug for KeyAuthorization {
 pub struct RetryPolicy {
     tries: u8,
     delay: Duration,
+    backpressure_factor: f32,
```
djc, line 499: I don't think this is a good name. Just call it `backoff`?

tests/pebble.rs
```diff
@@ -63,6 +63,28 @@ async fn http_01() -> Result<(), Box<dyn StdError>> {
         .map(|_| ())
 }
 
+#[tokio::test]
+#[ignore]
+async fn retry_policy() -> Result<(), Box<dyn StdError>> {
```
djc, line 68: It doesn't seem like this test is testing anything meaningful?

tests/pebble.rs
```diff
@@ -783,6 +805,7 @@ struct EnvironmentConfig {
     ///
     /// See <https://github.com/letsencrypt/pebble?tab=readme-ov-file#valid-authorization-reuse>
     authz_reuse: u8,
+    retry_policy: RetryPolicy,
```
djc, line 808: If we're going to do this, it should be in a separate commit, and we should just delete the `RETRY_POLICY` `const` in favor of the `EnvironmentConfig::default()` impl.

PR #119, opened 2025-06-30
Yield Break from RetryState::wait() on last iteration

djc:
As suggested by @christianhoelzl in https://github.com/djc/instant-acme/pull/116#discussion_r2174781560:

> One option is to preemptively end the retry if the next iteration would exceed the timeout. This is how [acme4j ](https://github.com/shred/acme4j/blob/2a5df329bdc34165f0f8f2c51c75a643e07aabb6/acme4j-client/src/main/java/org/shredzone/acme4j/PollableResource.java#L95)does it.

Doesn't really solve the original problem I highlighted, let's discuss here:

> Uhh, not sure how this interaction [between timeout and retry-after] will work? Seems like a strong chance the server might yield a Retry-After of more than the timeout set here, which means we'll only ever try once...

### DIFF
--- a/src/order.rs
+++ b/src/order.rs
@@ -164,13 +164,14 @@ impl Order {
 
     /// Poll the order with the given [`RetryPolicy`]
     ///
-    /// Yields the [`OrderStatus`] immediately if `Ready` or `Invalid`, or after `tries` attempts.
+    /// Yields the [`OrderStatus`] immediately if `Ready` or `Invalid`, or yields an
+    /// [`Error::Timeout`] if the [`RetryPolicy::timeout`] has been reached.
     pub async fn poll(&mut self, retries: &RetryPolicy) -> Result<OrderStatus, Error> {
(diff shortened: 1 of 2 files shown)


djc:
> That problem seems more in scope of the application that uses instant-acme. Not to retry only once, the application could enter a pending state and retry an certificate enrollment from scratch not before the given Retry-After date-time. For example if the CA does not issue certs the next days because you requested too many recently. If the application can observe the Retry-After value with the instant-acme api, then an application has the data at hand to handle this temporary error condition. Maybe extend `types::Error::Api` to yield the Retry-After header?

Yielding a specific error variant seems like a decent potential way out of this problem...

@cpu thoughts?

cpu:
> Yielding a specific error variant seems like a decent potential way out of this problem...

That sounds reasonable to me :+1: I agree with @christianhoelzl that this feels like a place where higher level application logic should take-over.

PR #120, opened 2025-06-30
Add support for loading existing accounts from key

### LINKED ISSUE #93: Account should be able to be created from a key only

djc:
Obsoletes #94, fixes #93.

cc @axos88

### DIFF
--- a/src/account.rs
+++ b/src/account.rs
@@ -6,6 +6,7 @@ use http::header::LOCATION;
 use http::{Method, Request};
 #[cfg(feature = "time")]
 use http_body_util::Full;
+use rustls_pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
@@ -214,7 +215,7 @@ impl Account {
 
         Ok(AccountCredentials {
             id: self.inner.id.clone(),
-            key_pkcs8: new_key_pkcs8.as_ref().to_vec(),
+            key_pkcs8: new_key_pkcs8,
             directory,
             urls,
         })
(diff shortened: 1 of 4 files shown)


src/account.rs
```diff
@@ -403,6 +403,29 @@ impl AccountBuilder {
...
+        server_url: String,
+    ) -> Result<(Account, AccountCredentials), Error> {
+        Self::create_inner(
+            &NewAccount {
+                contact: &[],
+                terms_of_service_agreed: true,
+                only_return_existing: true,
+            },
+            key,
+            None,
```
djc, line 424: My sense is that providing an external account binding does not make sense in the context of loading an existing key, but I didn't find anything explicit in RFC 8555 either way.

cpu: I agree the spec leaves EAB interaction with the non-creation newAccount requests (e.g. info update, only return existing to fetch the account URL) underspec'd.

IIRC I wasn't super jazzed about EAB making it into RFC 8555 because there were ~0 real implementations before the text was finalized since it wasn't a feature LE needed. Rough consensus and running code just gave us rough consensus this time I guess :P 

Pebble gained an implementation later on. Looking at it now it appears to only verify EAB after handling `onlyReturnExisting`, but I think will verify it for account updates: https://github.com/letsencrypt/pebble/blob/b656642cf581b37744ebf8d63c2e3c8358afae10/wfe/wfe.go#L1340-L1363 I'm not sure if there are any other open source implementations to peek at.

TL;DR - it feels reasonable to leave out until someone shows up with a concrete need :shrug:

cpu (approved):

src/account.rs
djc: Thanks, that's useful backstory!

PR #124, opened 2025-07-03
Add Account::key_thumbprint() for convenience

rickhan:
In my case, I need to get account key thumb, for quick http-01 challenge （for Lets Encrypt）
here is my example

```rust

#[get("/.well-known/acme-challenge/{token}")]
async fn handle_acme_challenge(token: web::Path<String>) -> impl Responder {
      let thumb = ACCOUNT.key_thumbprint();  // ACCOUNT is an instance of instant-acme::Account
      HttpResponseBuilder::new(StatusCode::OK)
        .content_type("text/plain")
        .body(format!(
            "{}.{}",
            token.into_inner(),
            thumb
        ))
}
```

### DIFF
--- a/src/account.rs
+++ b/src/account.rs
@@ -297,7 +297,7 @@ impl Account {
         &self.inner.id
     }
 
-    /// Get the account key_thumbprint
+    /// Get the [RFC 7638](https://www.rfc-editor.org/rfc/rfc7638) account key thumbprint
     pub fn key_thumbprint(&self) -> &str {
         &self.inner.key.thumb
     }


djc (approved):
I guess this is okay. @cpu thoughts?

src/account.rs
```diff
@@ -296,6 +296,11 @@ impl Account {
     pub fn id(&self) -> &str {
         &self.inner.id
     }
+
+    /// Get the account thumb
+    pub fn thumb(&self) -> &str {
```
djc, line 301: Maybe we should call it `key_thumbprint()`?

djc:
I think this is unrelated to the PR:

```
---- replacement_order stdout ----
Error: Api(Problem { type: Some("urn:ietf:params:acme:error:serverInternal"), detail: Some("could not find an order for the given certificate: could not find order resulting in the given certificate serial number"), status: Some(500), subproblems: [] })
```

It was happening on `main` yesterday, too. Not sure what's up with that...

PR #125, opened 2025-07-03
extend renewal_info() to yield retry-after header

christianhoelzl:
This pull request adds an extension to the function `Account::renewal_info()` to additionally yield the Retry-After header as a `SystemTime`. Reference is [RFC 9773, 4.3.2 Client Handling of Retry-After](https://www.rfc-editor.org/rfc/rfc9773.html#section-4.3.2). The `SystemTime` is a hint from the ACME server when to again fetch the renewal window.

### DIFF
--- a/src/account.rs
+++ b/src/account.rs
@@ -1,4 +1,6 @@
 use std::sync::Arc;
+#[cfg(feature = "time")]
+use std::time::{Duration, SystemTime};
 
 use base64::prelude::{BASE64_URL_SAFE_NO_PAD, Engine};
 use http::header::LOCATION;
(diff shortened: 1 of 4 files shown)


djc:
I think we should yield a Duration rather than a SystemTime.

christianhoelzl:
Changed to Duration. The edge case that the retry-after is in past is treated as a duration with 0 secs.
Could consider to move the fn `retry-after` to lib.rs?

djc:
> Could consider to move the fn retry-after to lib.rs?

Yes, please do that in a separate commit!

src/account.rs
```diff
@@ -166,7 +170,12 @@ impl Account {
             .body(Full::default())?;
 
         let rsp = self.inner.client.http.request(request).await?;
-        Problem::check::<RenewalInfo>(rsp).await
+        let retry_after = crate::order::retry_after(&rsp)
+            .ok_or(Error::Str("Missing Retry-After header"))?
+            .duration_since(std::time::SystemTime::now())
+            .unwrap_or(Duration::ZERO);
```
djc, line 176: Please rewrite it more like this:

```rust
let Some(retry_after) = retry_after(&rsp) else {
     return Err(Error::Str("missing Retry-After header"));
};

let delay = retry_after.duration_since(SystemTime::now()).unwrap_or(Duration::ZERO);
```

Import all the paths and keep the error message lowercase.

PR #126, opened 2025-07-03
introduce send_device_attestation() for device-attest-01 challenge

christianhoelzl:
This commit adds a new function `Order::send_device_attestation()`. This function can be used for a device-attest-01 challenge, that requires a payload to be provided when the ACME client notifies the ACME server that the challenge is ready. The challenge is described in the [draft-acme-device-attest](https://datatracker.ietf.org/doc/draft-acme-device-attest/). Pebble has no support for this, so it is untested so far in the pebble test.

This new API method allows to use the ACME challenge device-attest-01. Motivation to use this challenge is to improve device management based on the methods described in the link.

### DIFF
--- a/src/order.rs
+++ b/src/order.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::ops::{ControlFlow, Deref};
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime};
@@ -12,7 +13,7 @@ use tokio::time::sleep;
 use crate::account::AccountInner;
 use crate::types::{
     Authorization, AuthorizationState, AuthorizationStatus, AuthorizedIdentifier, Challenge,
-    ChallengeType, Empty, FinalizeRequest, OrderState, OrderStatus, Problem,
+    ChallengeType, DeviceAttestation, Empty, FinalizeRequest, OrderState, OrderStatus, Problem,
 };
 use crate::{Error, Key, crypto, nonce_from_response, retry_after};
 
@@ -443,6 +444,40 @@ impl ChallengeHandle<'_> {
         let rsp = self
             .account
             .post(Some(&Empty {}), self.nonce.take(), &self.challenge.url)
+            .await?;
+
+        *self.nonce = nonce_from_response(&rsp);
+        let response = Problem::check::<Challenge>(rsp).await?;
+        match response.error {
+            Some(details) => Err(Error::Api(details)),
+            None => Ok(()),
+        }
+    }
+
+    /// Notify the server that the challenge is ready by sending a device attestation
+    ///
+    /// This function is for the ACME challenge device-attest-01. It should not be used
+    /// with other challenge types.
+    /// See <https://datatracker.ietf.org/doc/draft-acme-device-attest/> for details.
+    ///
+    /// `payload` is the device attestation object as defined in link. Provide the attestation
+    /// object as a raw blob. Base64 encoding of the attestation object `payload.att_obj`
+    /// is done by this function.
+    pub async fn send_device_attestation(
+        &mut self,
+        payload: &DeviceAttestation<'_>,
+    ) -> Result<(), Error> {
+        if self.challenge.r#type != ChallengeType::DeviceAttest01 {
+            return Err(Error::Str("challenge type should be device-attest-01"));
+        }
+
+        let payload = DeviceAttestation {
+            att_obj: Cow::Owned(BASE64_URL_SAFE_NO_PAD.encode(&payload.att_obj).into()),
+        };
+
+        let rsp = self
+            .account
+            .post(Some(&payload), self.nonce.take(), &self.challenge.url)
             .await?;
 
         *self.nonce = nonce_from_response(&rsp);
--- a/src/types.rs
+++ b/src/types.rs
@@ -648,7 +648,7 @@ pub enum AuthorizationStatus {
 #[allow(missing_docs)]
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 #[non_exhaustive]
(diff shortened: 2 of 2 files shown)


src/order.rs
```diff
@@ -456,6 +456,28 @@ impl ChallengeHandle<'_> {
...
 
+    /// Notify the server that the challenge is ready and provide a payload
+    ///
+    /// This function is for the ACME challenge device-attest-01.
+    /// See <https://datatracker.ietf.org/doc/draft-acme-device-attest/> for details.
+    ///
+    /// Note: Do not use this function for http-01, tls-alpn-01 or dns-01 challenges.
+    ///
+    /// `payload` should serialize to "attObj" as defined in link.
+    pub async fn set_ready_with_payload(&mut self, payload: &impl Serialize) -> Result<(), Error> {
```
djc, line 467: Should we add a `ChallengeType` and error if the `ChallengeType` is not the expected one?

I suggest we take a more specific type here, maybe:

```rust
#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
struct DeviceAttestation<T> {
     att_obj: T,
}
```

and then take `object: &DeviceAttestation<impl Serialize>`?

Maybe rename to `send_device_attestation()`?

christianhoelzl:
This is an interesting discussion.

I also feel that it still is in an early stage. On the other hand, there already is support for this ACME challenge device-attest-01 in [Step CA](https://smallstep.com/blog/managed-device-attestation/) and maybe others. 

Regarding the API, if limiting the new challenge set ready function to device attestation, then the `&impl Serialize` is not required and a blob fits better to even do the base64 encoding of `payload.attObj`. I changed it this way in the last commit. 

If you look into the [acme-device-attest demo](https://github.com/brandonweeks/acme-device-attest-demo) from Brandon Weeks, the ACME server is Step CA and the ACME client based on a GO library from Google. This library provids an arbitrary payload for newer challenges, look [here](https://cs.opensource.google/go/x/crypto/+/master:acme/types.go;drc=b3790b8d914304c8187dc2c86800101c329d77cd;l=536). So I think long term the API could end up having both, challenge-specific and a generic function to add an arbitrary payload for other challenges beyond device-attest-01.

christianhoelzl:
Rebased onto main

src/types.rs
```diff
@@ -933,6 +935,14 @@ pub(crate) enum SigningAlgorithm {
 #[derive(Debug, Serialize)]
 pub(crate) struct Empty {}
 
+/// Attestation payload used for device-attest-01
```
djc, line 938: Nit: move this just above `Empty`?

djc (approved):
This seems sensible to me.

src/types.rs
```diff
@@ -930,6 +932,14 @@ pub(crate) enum SigningAlgorithm {
     Hs256,
 }
 
+/// Attestation payload used for device-attest-01
```
cpu, line 935: nit: I think it's worth a doc citation here too:

```suggestion
/// Attestation payload used for device-attest-01
///
/// See <https://datatracker.ietf.org/doc/draft-acme-device-attest/> for details.
```

cpu (commented):
What about the two new `Identifier` types? 

AIUI these can be included in a CSR. It would be fine to ignore them for the `rcgen` CSR generation code that this crate offers as a helper, but I think in all cases the CSR's identifiers need to match the order's identifiers, so an order would have to be constructed with the new identifier types even if you gin up a CSR externally, right?

src/order.rs
```diff
@@ -453,6 +454,42 @@ impl ChallengeHandle<'_> {
         }
     }
 
+    /// Notify the server that the challenge is ready and send device attestation
```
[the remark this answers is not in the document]
christianhoelzl: changed

src/order.rs
```diff
@@ -453,6 +454,42 @@ impl ChallengeHandle<'_> {
...
+    ///
+    /// This function is for the ACME challenge device-attest-01.
+    /// See <https://datatracker.ietf.org/doc/draft-acme-device-attest/> for details.
+    ///
+    /// Note: Do not use this function for http-01, tls-alpn-01 or dns-01 challenges.
```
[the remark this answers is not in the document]
christianhoelzl: changed to your suggestion

christianhoelzl:
I rebased and fixed the review comments (two new identifiers, better description). Thanks for the suggestions.

djc (approved):

src/types.rs
```diff
@@ -648,14 +648,24 @@ pub enum AuthorizationStatus {
...
+
+    /// Permanent Identifier
+    ///
+    /// Note that this identifier is only used for attestation.
+    PermanentIdentifier(String),
+
+    /// Hardware Module identifier
+    ///
+    /// Note that this identifier is only used for attestation.
+    HardwareModule(String),
```
cpu, line 668: I went down a bit of a rabbit hole thinking about what the value of these identifier types should be. I think `String` is the right choice for now, but I also filed some issues upstream about the spec because I think as written it leaves some important details up in the air w.r.t how CSRs and order identifiers interact:

https://github.com/brandonweeks/draft-bweeks-acme-device-attest/issues/9
https://github.com/brandonweeks/draft-bweeks-acme-device-attest/issues/10

cpu (approved):
Thanks! 

One last thought: have you tested this end-to-end with a server implementation? e.g. maybe that fork of `step-ca` you found? I think it'd be nice to confirm this works in practice before we merge it. I'm still a little bit nervous there's no test coverage.

src/types.rs
christianhoelzl: Thats great! If it turns out, that there is something better than `String` then I'll provide a pr.

cpu:
> I have partially tested with a private ACME test server such that the new method works fine

I would be OK merging on the basis of this testing if @djc concurs

PR #130, opened 2025-07-04
ci: add external types and semver compat checks

cpu:
The external types situation isn't great ATM (mostly due to the `HttpClient` trait, and the `Error` enum) but we can at least prevent backsliding with a CI check while considering improvements to reduce the number of blessed deps.


djc (approved):

djc:
Nice, thanks!

PR #132, opened 2025-07-04
Limit external types

djc:
I think this is probably worth it? The remaining ones are all 1.x+.


cpu (approved):
Nice improvements. :+1: Thank you.

Issue #135, opened 2025-07-13
ZeroSSL failure due to null `replaces` field


styris-ame:
When using ZeroSSL and attempting to call `new_order`, the following error occurs:

> Error: API error: The "replaces" field does not contain a '.' character. (urn:ietf:params:acme:error:malformed)

This is due to an unexpected null value on the `replaces` property, which ZeroSSL doesn't expect:
`{"replaces":null,"identifiers":[{"type":"dns","value":"zerossl-test.styris.net"}],"profile":null}`

Adding a `skip_serializing_if` attribute on `replaces` solves this:
```rust
#[derive(Debug, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct NewOrder<'a> {
    /// The [`CertificateIdentifier`] of a previously issued certificate being replaced by the order
    #[serde(skip_serializing_if = "Option::is_none")]
    pub(crate) replaces: Option<CertificateIdentifier<'a>>,
    /// Identifiers to be included in the order
    identifiers: &'a [Identifier],
    profile: Option<&'a str>,
}
```
Now when `replaces` is `None`, it is simply not serialized, solving the issue with ZeroSSL:
`{"identifiers":[{"type":"dns","value":"zerossl-test.styris.net"}],"profile":null}`

djc:
Thanks for reporting, want to send a PR?

PR #136, opened 2025-07-14
Skip serializing `replaces` field if None

styris-ame:
Fixes https://github.com/djc/instant-acme/issues/135)


djc:
Thanks! Preparing release in #137.

cpu:
Are there other places we should be giving the same treatment? E.g. in #135 @styris-ame said:
> Now when replaces is None, it is simply not serialized, solving the issue with ZeroSSL:
{"identifiers":[{"type":"dns","value":"zerossl-test.styris.net"}],"profile":null}

Should `profile` be serialized when `null`?

cpu:
> Should profile be serialized when null?

I did a quick pass and this seems to be the only other `Optional` field without the annotation in a `Serializable` type we send to a peer. I'll open a quick fix PR in a sec.

PR #137, opened 2025-07-14
Prepare 0.8.1

djc:
## Proposed release notes

- Avoid encoding `replacing: null` in new order requests for ZeroSSL compatibility (see #136).

For #136.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,7 +681,7 @@ dependencies = [
 
 [[package]]
 name = "instant-acme"
-version = "0.8.0"
+version = "0.8.1"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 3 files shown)


christianhoelzl:
When do you plan to release 0.8.1? I started testing the experimental attestation feature with Step CA and could provide a fix tomorrow in case there is a time window for this.

cpu:
@christianhoelzl I think it's easy to make a 0.8.2 when you're ready, no need to rush :-)

cpu (approved):
Thanks! It seems reasonable to omit a mention of the profiles encoding change in the release notes since unlike the `replaces` change it's not fixing a known ACME server incompatibility and shouldn't be user facing.

djc:
* Published instant-acme v0.8.1 at registry `crates-io`
* [new tag]           0.8.1 -> 0.8.1
* [Release notes](https://github.com/djc/instant-acme/releases/tag/0.8.1)

PR #145, opened 2025-07-14
Add `USER_AGENT` header to requests

styris-ame:
Fixes https://github.com/djc/instant-acme/issues/144

Note that this PR also renames the `req` variable to `request` in `Client::new` for consistency.

Perhaps the user agent could be made to be configurable externally, however I see no real need for that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@ use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
 use bytes::{Buf, Bytes};
-use http::header::{CONTENT_TYPE, RETRY_AFTER};
+use http::header::{CONTENT_TYPE, RETRY_AFTER, USER_AGENT};
 use http::{Method, Request, Response, StatusCode};
 use http_body::{Frame, SizeHint};
 use http_body_util::BodyExt;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -53,6 +53,8 @@ pub use types::{
 };
 use types::{Directory, JoseJson, Signer};
 
+const CRATE_USER_AGENT: &str = concat!("instant_acme/", env!("CARGO_PKG_VERSION"));
```
djc, line 56: Please move this down to somewhere near the bottom of the module (though before `mod tests {}`), to match the top-down style of the crate.

Also, please spell the crate name with a hyphen instead of an underscore.

djc (approved):
Thanks! How many servers are you testing against? Should we hold off on the next release until you're done testing?

(Nit: mentioning an issue in a commit message tends to get a little noisy if you push rebased or amended versions of the same commit over time, better to keep that constrained to the PR description. Also feel free to submit PRs without filing an issue first for stuff like this!)

styris-ame:
> (Nit: mentioning an issue in a commit message tends to get a little noisy if you push rebased or amended versions of the same commit over time, better to keep that constrained to the PR description. Also feel free to submit PRs without filing an issue first for stuff like this!)

Thanks for the pointers! I'll keep that in mind.

djc (approved):

djc:
If you want to send a PR with a version bump, we can get this puppy released!

PR #208, opened 2025-11-04
feat: add create_from_key for accounts with user-provided keys

ptrus:
Hi, Thanks for the great library! 
I needed a way to create ACME accounts using externally managed keys, so this PR adds that functionality. I made the change backwards compatible (i.e., introduced a new public method without changing the existing ones).

Allows creating ACME accounts using externally managed keys, instead of being limited to loading existing ones or registering with library-generated keys.

Please let me know if anything should be updated or improved.


djc (approved):
LGTM, thanks. @cpu any thoughts?

PR #226, opened 2025-12-17
Preserve the original order of identifiers when generating the CSR 

tarka:
Preserve the original order of identifiers when generating the CSR to match the behaviour of certbot, etc.

Currently the CSR generator uses the authorisations to populate the SAN; however letsencrypt sorts these alphabetically. letsencrypt also uses the first SAN as the `Subject: CN=` if none is explicitly provided. This means that if the user specifies an identifier list of `["www.example.com", "staging.example.com"]`, the resulting certificate will have `Subject: CN=staging.example.com`. certbot preserves the original ordering (and doesn't set the subject), and others appear to behave in a similar way (e.g. acme.sh).

The PR caches the original identifiers and uses it for the SAN generation (if available), or falls back to the current behaviour if they are not available (e.g. if an existing order is retrieved from the CA).

### DIFF
--- a/src/account.rs
+++ b/src/account.rs
@@ -129,8 +129,8 @@ impl Account {
             .map(|id| match id {
                 Identifier::Dns(name) => name.clone(),
                 Identifier::Ip(ip) => ip.to_string(),
-                Identifier::PermanentIdentifier(name) => name.to_string(),
-                Identifier::HardwareModule(name) => name.to_string(),
+                Identifier::PermanentIdentifier(name) => name.clone(),
+                Identifier::HardwareModule(name) => name.clone(),
             })
             .collect();
 


djc (commented):
I think this could make sense.

Just curious what problem this solves for you exactly?

tarka:
@djc The primary issue I had was that this is surprising behaviour, as the effective reference implementation (`certbot`) explicitly states that the first domain == Subject. However having read @cpu's comments I'm inclined to agree that this will be undefined behaviour in the future.

@cpu As mentioned, the purpose of this was to match the behaviour of most (all?) other ACME clients. However I wasn't aware that the subject field was removed for other LE profiles, and I agree 'classic' will probably be deprecated at some point, so I will have to work around this anyway.

Thanks for the discussion but I'm happy for you to close this.

PR #249, opened 2026-02-06
cargo-audit: ignore RUSTSEC-2026-0009

cpu:
Follow-up to https://github.com/djc/instant-acme/pull/248

Ignores [RUSTSEC-2026-0009](https://rustsec.org/advisories/RUSTSEC-2026-0009.html) for the reasons discussed in #248 and the associated comment in the `audit.toml` file. We can remove this with a future MSRV update to 1.88+.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,5 +1,11 @@
 [advisories]
 yanked = "deny"
+ignore = [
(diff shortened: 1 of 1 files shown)


djc:
Yes, needs a change in `deny.toml`.

PR #254, opened 2026-02-12
fix: send_device_attestation yields challenge status

christianhoelzl:
The pull request extends the function `send_device_attestation`. It yields the `ChallengeStatus` instead of an empty success result. This is useful to check if the attestation object was validated by the ACME CA immediately after sending the attestation object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ If both `ring` and `aws-lc-rs` are enabled, `aws-lc-rs` will be used.
 ## Limitations
 
 * Only supports P-256 ECDSA account keys for now
+* Device attestation support is experimental
 
 ## Getting started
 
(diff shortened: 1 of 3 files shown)


djc (approved):
LGTM -- is this from a spec change, or just a different kind of testing?

djc:
Okay. I think it would be good to add some testing for this stuff. In an ideal world, it would be nice to get support for this draft into Pebble -- do you think that's something that could be in scope for you?

djc:
Forking Pebble doesn't seem appealing, I don't think adding a custom Rust-based ACME server would help if we build it.

Would be good if you can file an issue against Pebble to at least ask if they'd be open to supporting it (and linking that here).

cpu (approved):
As mentioned in https://github.com/djc/instant-acme/pull/255/changes#r2799709957 I think this is semver breaking, but I don't think it's worth bumping our minor version or leaving it broken. I suspect adoption is very low and provided we mark this stuff experimental it feels like the right-trade off to ignore semver and fix it.

cpu (approved):
Thanks, LGTM mod the `rust fmt` CI failure.

PR #265, opened 2026-02-23
Support challenge types without tokens (including dns-persist-01)

djc:
Working towards fixing the tests/including dns-persist-01 support (not sure yet if the latter is required for the former).

Gotta stop now, if you want to take a look, feel free to take the baton?

### DIFF
--- a/examples/provision.rs
+++ b/examples/provision.rs
@@ -71,7 +71,7 @@ async fn main() -> anyhow::Result<()> {
         println!(
             "_acme-challenge.{} IN TXT {}",
             challenge.identifier(),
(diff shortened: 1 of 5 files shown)


cpu:
>  I want to get to this at some point soon if you don't beat me to it.

Here's an attempt at both completing the move away from having `token` in the `Challenge` regardless of type, and also adding end-to-end dns-persist-01 challenge support: https://github.com/djc/instant-acme/pull/279

cpu:
Let's run with https://github.com/djc/instant-acme/pull/283

Issue #267, opened 2026-02-25
Support for more crypto backends


jcgruenhage:
Motivated by f.ex. #266, it could make sense to introduce additional crypto backends. Candidates I have in mind:
- `openssl`: Is already in use everywhere anyway and supports everything under the sun, including more niche things like P-521 and Ed448
- RustCrypto: Might not make sense to use. Most of the crates have not seen formal audits, but they do have implementations for most niche things as well, contrary to `ring` and `aws-lc-rs`.

My assumption is that this is a no, but I wanted to make sure this assumption is correct instead of just assuming it blindly, which is why I'm opening an issue about it.

djc:
Sounds great, thanks!

cpu:
Resolved with https://github.com/djc/instant-acme/pull/268

PR #268, opened 2026-02-25
Allow library users to bring their own crypto providers

jcgruenhage:
Implements base for #267, allowing users to bring their own crypto.

Marked as draft for now, will un-draft once I've implemented (and tested) backends with other crypto libraries, like openssl and rustcrypto to validate this approach is viable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,10 @@ tracing-subscriber =  { version = "0.3.16", features = ["env-filter"] }
 name = "provision"
 required-features = ["hyper-rustls", "rcgen", "aws-lc-rs"]
 
+[[example]]
+name = "certgen"
+required-features = ["rcgen", "aws-lc-rs"]
+
 # Pebble integration test.
 # Ignored by default because it requires pebble & pebble-challtestsrv.
 # Run with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ tracing-subscriber =  { version = "0.3.16", features = ["env-filter"] }
 
 [[example]]
 name = "provision"
-required-features = ["hyper-rustls", "rcgen"]
+required-features = ["hyper-rustls", "rcgen", "aws-lc-rs"]
 
 # Pebble integration test.
 # Ignored by default because it requires pebble & pebble-challtestsrv.
@@ -63,7 +63,7 @@ required-features = ["hyper-rustls", "rcgen"]
 #  PEBBLE=path/to/pebble CHALLTESTSRV=path/to/pebble-challtestsrv cargo test -- --ignored
 [[test]]
 name = "pebble"
-required-features = ["hyper-rustls"]
+required-features = ["hyper-rustls", "rcgen"]
 
 [package.metadata.docs.rs]
 # all non-default features except fips (cannot build on docs.rs environment)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "instant-acme"
-version = "0.8.5"
+version = "0.9.0"
 edition = "2021"
 rust-version = "1.70"
 license = "Apache-2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ allowed_external_types = [
     "http::*",
     "http_body::*",
     "hyper::*",
+    "rustls::crypto::CryptoProvider",
     "rustls_pki_types::*",
     "serde_core::*",
     "serde_json::*",
--- a/src/account.rs
+++ b/src/account.rs
@@ -235,7 +235,7 @@ impl Account {
             old_key: Jwk<'a>,
         }
 
-        let (new_key, new_key_pkcs8) = Key::generate_pkcs8()?;
+        let (new_key, new_key_pkcs8) = Key::generate()?;
         let mut header = new_key.header(Some("nonce"), new_key_url);
         header.nonce = None;
         let payload = NewKey {
@@ -440,7 +440,7 @@ impl AccountBuilder {
         directory_url: String,
         external_account: Option<&ExternalAccountKey>,
     ) -> Result<(Account, AccountCredentials), Error> {
-        let (key, key_pkcs8) = Key::generate_pkcs8()?;
+        let (key, key_pkcs8) = Key::generate()?;
         Self::create_inner(
             account,
             (key, key_pkcs8),
@@ -590,14 +590,7 @@ pub struct Key {
 
 impl Key {
     /// Generate a new ECDSA P-256 key pair
-    #[deprecated(since = "0.8.3", note = "use `generate_pkcs8()` instead")]
-    pub fn generate() -> Result<(Self, PrivateKeyDer<'static>), Error> {
-        let (key, pkcs8) = Self::generate_pkcs8()?;
-        Ok((key, PrivateKeyDer::Pkcs8(pkcs8)))
-    }
-
-    /// Generate a new ECDSA P-256 key pair
-    pub fn generate_pkcs8() -> Result<(Self, PrivatePkcs8KeyDer<'static>), Error> {
+    pub fn generate() -> Result<(Self, PrivatePkcs8KeyDer<'static>), Error> {
         let rng = crypto::SystemRandom::new();
         let pkcs8 =
             crypto::EcdsaKeyPair::generate_pkcs8(&crypto::ECDSA_P256_SHA256_FIXED_SIGNING, &rng)
--- a/src/account.rs
+++ b/src/account.rs
@@ -232,7 +232,7 @@ impl Account {
         struct NewKey<'a> {
             account: &'a str,
             #[serde(rename = "oldKey")]
-            old_key: Jwk,
+            old_key: Jwk<'a>,
         }
 
         let (new_key, new_key_pkcs8) = Key::generate_pkcs8()?;
@@ -617,7 +617,7 @@ impl Key {
 
     fn new(pkcs8_der: &[u8], rng: crypto::SystemRandom) -> Result<Self, Error> {
         let inner = crypto::p256_key_pair_from_pkcs8(pkcs8_der, &rng)?;
-        let thumb = BASE64_URL_SAFE_NO_PAD.encode(Jwk::thumb_sha256(&inner)?);
+        let thumb = BASE64_URL_SAFE_NO_PAD.encode(Jwk::new(&inner).thumb_sha256()?);
         Ok(Self {
             rng,
             signing_algorithm: SigningAlgorithm::Es256,
--- a/src/account.rs
+++ b/src/account.rs
@@ -13,6 +13,8 @@ use http::{Method, Request};
 #[cfg(feature = "hyper-rustls")]
 use rustls::RootCertStore;
 #[cfg(feature = "hyper-rustls")]
+use rustls::crypto::CryptoProvider;
+#[cfg(feature = "hyper-rustls")]
 use rustls_pki_types::CertificateDer;
 #[cfg(feature = "hyper-rustls")]
 use rustls_pki_types::pem::PemObject;
@@ -52,9 +54,9 @@ pub struct Account {
 impl Account {
     /// Create an account builder with the default HTTP client
     #[cfg(feature = "hyper-rustls")]
-    pub fn builder() -> Result<AccountBuilder, Error> {
+    pub fn builder(rustls_crypto_provider: CryptoProvider) -> Result<AccountBuilder, Error> {
         Ok(AccountBuilder {
-            http: Box::new(DefaultClient::try_new()?),
+            http: Box::new(DefaultClient::try_new(rustls_crypto_provider)?),
         })
     }
 
@@ -63,7 +65,10 @@ impl Account {
     /// This is useful if your ACME server uses a testing PKI and not a certificate
     /// chain issued by a publicly trusted CA.
     #[cfg(feature = "hyper-rustls")]
-    pub fn builder_with_root(pem_path: impl AsRef<Path>) -> Result<AccountBuilder, Error> {
+    pub fn builder_with_root(
+        pem_path: impl AsRef<Path>,
+        rustls_crypto_provider: CryptoProvider,
+    ) -> Result<AccountBuilder, Error> {
         let root_der = match CertificateDer::from_pem_file(pem_path) {
             Ok(root_der) => root_der,
             Err(err) => return Err(Error::Other(err.into())),
@@ -72,7 +77,7 @@ impl Account {
         let mut roots = RootCertStore::empty();
         match roots.add(root_der) {
             Ok(()) => Ok(AccountBuilder {
-                http: Box::new(DefaultClient::with_roots(roots)?),
+                http: Box::new(DefaultClient::with_roots(roots, rustls_crypto_provider)?),
             }),
             Err(err) => Err(Error::Other(err.into())),
         }
--- a/src/account.rs
+++ b/src/account.rs
@@ -1,6 +1,7 @@
 #[cfg(feature = "hyper-rustls")]
 use std::path::Path;
 use std::sync::Arc;
+
 #[cfg(feature = "time")]
 use std::time::{Duration, SystemTime};
 
@@ -11,30 +12,29 @@ use http::header::USER_AGENT;
 #[cfg(feature = "time")]
 use http::{Method, Request};
 #[cfg(feature = "hyper-rustls")]
-use rustls::RootCertStore;
+use rustls::{RootCertStore, crypto::CryptoProvider as RustlsCryptoProvider};
 #[cfg(feature = "hyper-rustls")]
-use rustls::crypto::CryptoProvider;
-#[cfg(feature = "hyper-rustls")]
-use rustls_pki_types::CertificateDer;
-#[cfg(feature = "hyper-rustls")]
-use rustls_pki_types::pem::PemObject;
+use rustls_pki_types::{CertificateDer, pem::PemObject};
 use rustls_pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
 #[cfg(feature = "hyper-rustls")]
 use crate::DefaultClient;
+use crate::crypto::HmacKey;
+#[cfg(doc)]
+use crate::crypto::HmacKeyProvider;
+use crate::crypto::{CryptoProvider, SigningKey};
 use crate::order::Order;
 use crate::types::{
     AccountCredentials, AuthorizationStatus, Empty, Header, JoseJson, Jwk, KeyOrKeyId, NewAccount,
     NewAccountPayload, NewOrder, OrderState, Problem, ProfileMeta, RevocationRequest, Signer,
-    SigningAlgorithm,
 };
 #[cfg(feature = "time")]
 use crate::types::{CertificateIdentifier, RenewalInfo};
 #[cfg(feature = "time")]
 use crate::{BodyWrapper, CRATE_USER_AGENT, retry_after};
-use crate::{BytesResponse, Client, Error, HttpClient, crypto, nonce_from_response};
+use crate::{BytesResponse, Client, Error, HttpClient, nonce_from_response};
 
 /// An ACME account as described in RFC 8555 (section 7.1.2)
 ///
@@ -52,11 +52,18 @@ pub struct Account {
 }
 
 impl Account {
-    /// Create an account builder with the default HTTP client
+    /// Create an account builder with the given [`CryptoProvider`] and default HTTP client
+    ///
+    /// The `rustls_crypto_provider` configures TLS for the HTTPS connection to the
+    /// ACME server and is separate from the ACME-level [`CryptoProvider`].
     #[cfg(feature = "hyper-rustls")]
-    pub fn builder(rustls_crypto_provider: CryptoProvider) -> Result<AccountBuilder, Error> {
+    pub fn builder(
+        provider: &'static CryptoProvider,
+        rustls_crypto_provider: RustlsCryptoProvider,
+    ) -> Result<AccountBuilder, Error> {
         Ok(AccountBuilder {
             http: Box::new(DefaultClient::try_new(rustls_crypto_provider)?),
+            provider,
         })
     }
 
@@ -67,7 +74,8 @@ impl Account {
     #[cfg(feature = "hyper-rustls")]
     pub fn builder_with_root(
         pem_path: impl AsRef<Path>,
-        rustls_crypto_provider: CryptoProvider,
+        provider: &'static CryptoProvider,
+        rustls_crypto_provider: RustlsCryptoProvider,
     ) -> Result<AccountBuilder, Error> {
         let root_der = match CertificateDer::from_pem_file(pem_path) {
             Ok(root_der) => root_der,
@@ -78,14 +86,18 @@ impl Account {
         match roots.add(root_der) {
             Ok(()) => Ok(AccountBuilder {
                 http: Box::new(DefaultClient::with_roots(roots, rustls_crypto_provider)?),
+                provider,
             }),
             Err(err) => Err(Error::Other(err.into())),
         }
     }
 
-    /// Create an account builder with the given HTTP client
-    pub fn builder_with_http(http: Box<dyn HttpClient>) -> AccountBuilder {
-        AccountBuilder { http }
+    /// Create an account builder with the given [`CryptoProvider`] and HTTP client
+    pub fn builder_with_http(
+        http: Box<dyn HttpClient>,
+        provider: &'static CryptoProvider,
+    ) -> AccountBuilder {
+        AccountBuilder { http, provider }
     }
 
     /// Create a new order based on the given [`NewOrder`]
@@ -235,12 +247,12 @@ impl Account {
             old_key: Jwk<'a>,
         }
 
-        let (new_key, new_key_pkcs8) = Key::generate()?;
+        let (new_key, new_key_pkcs8) = Key::generate(self.inner.key.provider)?;
         let mut header = new_key.header(Some("nonce"), new_key_url);
         header.nonce = None;
         let payload = NewKey {
             account: &self.inner.id,
-            old_key: Jwk::new(&self.inner.key.inner),
+            old_key: self.inner.key.inner.as_jwk(),
         };
 
         let body = JoseJson::new(Some(&payload), header, &new_key)?;
@@ -343,8 +355,8 @@ impl Account {
     }
 
     /// Get the [RFC 7638](https://www.rfc-editor.org/rfc/rfc7638) account key thumbprint
-    pub fn key_thumbprint(&self) -> &str {
-        &self.inner.key.thumb
+    pub fn key_thumbprint(&self) -> Result<String, Error> {
+        Ok(BASE64_URL_SAFE_NO_PAD.encode(self.inner.key.thumb_sha256()?))
     }
 }
 
@@ -358,10 +370,11 @@ impl AccountInner {
     async fn from_credentials(
         credentials: AccountCredentials,
         http: Box<dyn HttpClient>,
+        provider: &'static CryptoProvider,
     ) -> Result<Self, Error> {
         Ok(Self {
             id: credentials.id,
-            key: Key::from_pkcs8_der(credentials.key_pkcs8)?,
+            key: Key::from_pkcs8_der(credentials.key_pkcs8, provider)?,
             client: Arc::new(match (credentials.directory, credentials.urls) {
                 (Some(directory_url), _) => Client::new(directory_url, http).await?,
                 (None, Some(directory)) => Client {
@@ -395,12 +408,12 @@ impl AccountInner {
 }
 
 impl Signer for AccountInner {
-    type Signature = <Key as Signer>::Signature;
+    type Signature = Vec<u8>;
 
     fn header<'n, 'u: 'n, 's: 'u>(&'s self, nonce: Option<&'n str>, url: &'u str) -> Header<'n> {
         debug_assert!(nonce.is_some());
         Header {
-            alg: self.key.signing_algorithm,
+            alg: self.key.inner.jws_algorithm(),
             key: KeyOrKeyId::KeyId(&self.id),
             nonce,
             url,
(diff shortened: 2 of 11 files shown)


src/crypto.rs
```diff
@@ -0,0 +1,232 @@
...
+
+/// SHA-256 hash function for ACME protocol operations.
+///
+/// Used for JWK thumbprints ([RFC 7638]) and challenge digests ([RFC 8555]).
+///
+/// [RFC 7638]: https://www.rfc-editor.org/rfc/rfc7638
+/// [RFC 8555]: https://www.rfc-editor.org/rfc/rfc8555
+pub trait Sha256: Send + Sync {
+    /// Compute the SHA-256 digest of the given data.
+    fn hash(&self, data: &[u8]) -> Box<dyn AsRef<[u8]>>;
```
djc: Can we yield an array here instead of a boxed trait object?

src/crypto.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    impl HmacSha256 for BuiltinHmacSha256 {
+        fn sign(&self, key: &[u8], data: &[u8]) -> Box<dyn AsRef<[u8]>> {
+            let hmac_key = ring_like::hmac::Key::new(ring_like::hmac::HMAC_SHA256, key);
+            Box::new(ring_like::hmac::sign(&hmac_key, data))
+        }
+    }
+}
+
+impl CryptoProvider {
```
djc: I think we should drop `builtin()` and have specific constructors for `ring()` and `aws_lc()`.

src/order.rs
```diff
@@ -526,34 +527,41 @@ impl Deref for ChallengeHandle<'_> {
 /// Refer to the methods below to see which encoding to use for your challenge type.
 ///
 /// <https://datatracker.ietf.org/doc/html/rfc8555#section-8.1>
-pub struct KeyAuthorization(String);
+pub struct KeyAuthorization {
+    inner: String,
+    provider: &'static CryptoProvider,
```
djc: Suggest only storing the `&dyn Sha256` here.

djc (commented):
Here's some early feedback. I like the direction overall!

src/account.rs
```diff
@@ -647,36 +713,48 @@ impl Signer for Key {
...
 
 impl Signer for ExternalAccountKey {
-    type Signature = crypto::hmac::Tag;
+    type Signature = Vec<u8>;
 
     fn header<'n, 'u: 'n, 's: 'u>(&'s self, nonce: Option<&'n str>, url: &'u str) -> Header<'n> {
         debug_assert_eq!(nonce, None);
         Header {
-            alg: SigningAlgorithm::Hs256,
+            alg: "HS256",
```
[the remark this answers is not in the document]
jcgruenhage: If a crypto provider with another key type shows up, they can't put their algorithm in here.

src/types.rs
```diff
@@ -265,64 +264,11 @@ pub(crate) struct Header<'a> {
...
-
-#[derive(Debug, Serialize)]
-pub(crate) struct Jwk {
-    alg: SigningAlgorithm,
-    crv: &'static str,
-    kty: &'static str,
-    r#use: &'static str,
-    x: String,
-    y: String,
-}
```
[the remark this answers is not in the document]
jcgruenhage: I'll bring them back, it was a bit easier to test it that way. I commented them out mentioning that I'll bring them back.

src/types.rs
```diff
@@ -265,62 +265,87 @@ pub(crate) struct Header<'a> {
...
+/// Each variant's fields are declared in lexicographic order for correct
+/// [RFC 7638](https://www.rfc-editor.org/rfc/rfc7638) thumbprint computation.
 #[derive(Debug, Serialize)]
-struct JwkThumb<'a> {
-    crv: &'a str,
-    kty: &'a str,
-    x: &'a str,
-    y: &'a str,
+#[serde(untagged)]
+pub(crate) enum JwkInner<'a> {
```
djc: Nit: how about we name this something more like `KeyFields`?

src/crypto.rs
```diff
@@ -0,0 +1,310 @@
...
+            let hmac_key = ring::hmac::Key::new(ring::hmac::HMAC_SHA256, key);
+            ring::hmac::sign(&hmac_key, data)
+                .as_ref()
+                .try_into()
+                .expect("HMAC-SHA-256 output is always 32 bytes")
+        }
+    }
+}
+
+impl CryptoProvider {
```
djc: This `impl` should be below the type definition for `CryptoProvider` (and before any traits impls).

src/crypto.rs
```diff
@@ -0,0 +1,310 @@
...
+                r#use: "sig",
+            }
+        }
+    }
+
+    struct BuiltinSha256;
+
+    impl Sha256 for BuiltinSha256 {
+        fn hash(&self, data: &[u8]) -> [u8; 32] {
+            ring::digest::digest(&ring::digest::SHA256, data)
```
djc: Nit: can we skip the `ring::` prefixes inside the `backend_ring` module?

Cargo.toml
```diff
@@ -53,17 +53,21 @@ tokio = { version = "1.22.0", features = ["macros", "rt", "rt-multi-thread", "ti
 tracing = "0.1.37"
 tracing-subscriber =  { version = "0.3.16", features = ["env-filter"] }
 
+[[example]]
```
djc: Nit: this should be in a separate commit.

djc (commented):
Looking good, appreciate the clean commit history!

src/account.rs
```diff
@@ -50,20 +52,24 @@ pub struct Account {
...
-    #[cfg(feature = "hyper-rustls")]
-    pub fn builder() -> Result<AccountBuilder, Error> {
+    /// Create an account builder with the given [`CryptoProvider`] and default HTTP client
+    #[cfg(all(feature = "hyper-rustls", any(feature = "aws-lc-rs", feature = "ring")))]
+    pub fn builder(provider: &'static CryptoProvider) -> Result<AccountBuilder, Error> {
         Ok(AccountBuilder {
             http: Box::new(DefaultClient::try_new()?),
+            provider,
         })
     }
```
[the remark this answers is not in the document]
jcgruenhage: I think that's reasonable, but it's also a breaking change. I'll do that in a different commit.

src/types.rs
```diff
@@ -974,13 +974,25 @@ fn deserialize_static_certificate_identifier<'de, D: serde::Deserializer<'de>>(
...
 }
 
+/// Algorithm identifier for JWS headers
+///
+/// See the [IANA JOSE registry](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms)
+/// for the full list of registered algorithms.
 #[derive(Clone, Copy, Debug, Serialize)]
 #[serde(rename_all = "UPPERCASE")]
 pub(crate) enum SigningAlgorithm {
+#[non_exhaustive]
```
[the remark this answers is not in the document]
jcgruenhage: Was correct in the end, but the commit introducing this change placed it in the wrong location. Good catch!

src/types.rs
jcgruenhage: Given that they are all fields of the Json Web Key, just `KeyFields` feels a bit off, I went with `JwkThumbFields`, as it's specifically the fields required for JWK thumbprint calc

src/types.rs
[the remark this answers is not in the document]
jcgruenhage: Fixed now :)

Cargo.toml
jcgruenhage: Split out into 6b1f6f9

src/account.rs
```diff
@@ -63,7 +65,10 @@ impl Account {
     /// This is useful if your ACME server uses a testing PKI and not a certificate
     /// chain issued by a publicly trusted CA.
     #[cfg(feature = "hyper-rustls")]
-    pub fn builder_with_root(pem_path: impl AsRef<Path>) -> Result<AccountBuilder, Error> {
+    pub fn builder_with_root(
+        pem_path: impl AsRef<Path>,
+        rustls_crypto_provider: CryptoProvider,
```
cpu: nit: It feels a little more natural to me to have the `CryptoProvider` as the first arg, and the `pem_path` as the second. Am I an outlier there?

src/types.rs
```diff
@@ -265,62 +265,88 @@ pub(crate) struct Header<'a> {
...
+    /// The intended use (`"sig"` for signing)
     r#use: &'static str,
-    x: String,
-    y: String,
 }
 
-impl Jwk {
-    pub(crate) fn new(key: &crypto::EcdsaKeyPair) -> Self {
+impl Jwk<'_> {
+    pub(crate) fn new(key: &crypto::EcdsaKeyPair) -> Jwk<'_> {
```
cpu: Does this need to change from `Self` because of the lifetime?

src/account.rs
```diff
@@ -590,14 +590,7 @@ pub struct Key {
 
 impl Key {
     /// Generate a new ECDSA P-256 key pair
-    #[deprecated(since = "0.8.3", note = "use `generate_pkcs8()` instead")]
```
cpu, line 588: I'm a bit confused by this change. I thought the reason for deprecating `generate()` was the name was ambiguous w.r.t to the encoding, where-as `generate_pkcs8()` was explicit. If that's the correct framing, then this feels like a step backwards and I'd have expected `generate()` to be removed and all the call-sites switched to `generate_pkcs8()`.

@djc Am I mischaracterizing the original deprecation intent?

src/lib.rs
```diff
@@ -200,18 +204,23 @@ struct DefaultClient(HyperClient<hyper_rustls::HttpsConnector<HttpConnector>, Bo
...
-                .try_with_platform_verifier()
+                .with_provider_and_platform_verifier(rustls_crypto_provider)
                 .map_err(|e| Error::Other(Box::new(e)))?,
         ))
     }
 
-    fn with_roots(roots: rustls::RootCertStore) -> Result<Self, Error> {
+    fn with_roots(
+        roots: rustls::RootCertStore,
+        rustls_crypto_provider: CryptoProvider,
```
cpu: same feedback to consider here w.r.t ordering the crypto provider arg first.

src/crypto.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::Error;
+use crate::types::{Jwk, SigningAlgorithm};
+
+#[cfg(feature = "aws-lc-rs")]
+mod aws_lc_rs;
+
+#[cfg(feature = "ring")]
+mod ring;
+
+/// Cryptographic provider for ACME operations.
```
cpu: It feels like it might be worth a sentence or two describing how this differs from a Rustls cryptography provider used for the TLS operations required for the ACME server HTTPS API, as opposed to the crypto operations required for the ACME protocol. WDYT?

src/crypto.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// Use [`CryptoProvider::aws_lc_rs()`] or [`CryptoProvider::ring()`] for the built-in backend,
```
cpu: nit: 'for the built-in backend' -> 'for a built-in backend' ?

src/crypto.rs
```diff
@@ -0,0 +1,107 @@
...
+    ///
+    /// [RFC 7638]: https://www.rfc-editor.org/rfc/rfc7638
+    /// [RFC 8555 section 8.1]: https://www.rfc-editor.org/rfc/rfc8555#section-8.1
+    pub sha256: &'static dyn Sha256,
+    /// HMAC-SHA-256 for External Account Binding.
+    ///
+    /// See [RFC 8555 section 7.3.4].
+    ///
+    /// [RFC 8555 section 7.3.4]: https://www.rfc-editor.org/rfc/rfc8555#section-7.3.4
+    pub hmac_sha256: &'static dyn HmacSha256,
```
cpu: In theory someone might want to use HS384 or HS512 here and the spec allows any JWS registry alg that's an HMAC, but I don't know if that's actually common in practice. It feels OK to me to land this as-is and consider additional HMACs later, vs trying to land something more general without a concrete need.

src/crypto.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+/// Load existing account keys and generate new ones.
+///
+/// Implementations are backend-specific (ring, aws-lc-rs, openssl, etc.)
+/// and key-type-specific (P-256, Ed25519, RSA, etc.).
+pub trait SigningKeyProvider: Send + Sync {
+    /// Load a signing key from PKCS#8 DER encoding.
+    fn load_key(&self, pkcs8: PrivatePkcs8KeyDer<'static>) -> Result<Box<dyn SigningKey>, Error>;
```
cpu, line 69: I'm slightly inclined to think we should name this 'load_pkcs8_key()`. How do others feel?

src/crypto.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// Generate a new key pair, returning the key and its PKCS#8 DER encoding.
+    fn generate_key(&self) -> Result<(Box<dyn SigningKey>, PrivatePkcs8KeyDer<'static>), Error>;
```
cpu, line 72: ditto here RE: naming to capture `pkcs8`. It's information carried by the `PrivatePkcs8KeyDer`, but feels worth considering a more explicit name.

src/account.rs
```diff
@@ -52,11 +50,15 @@ pub struct Account {
 }
 
 impl Account {
-    /// Create an account builder with the default HTTP client
+    /// Create an account builder with the given [`CryptoProvider`] and default HTTP client
```
cpu, line 55: Should the `rustls_crypto_provider` arg be mentioned here?

src/account.rs
```diff
@@ -52,11 +50,15 @@ pub struct Account {
...
     #[cfg(feature = "hyper-rustls")]
-    pub fn builder(rustls_crypto_provider: CryptoProvider) -> Result<AccountBuilder, Error> {
+    pub fn builder(
+        provider: &'static CryptoProvider,
+        rustls_crypto_provider: RustlsCryptoProvider,
```
cpu, line 62: I'm not sure I love how we require an explicit `RustlsCryptoProvider` arg here, but I know djc requested it previously. To me it feels like we're maybe being inconsistent with the API and defaults. 

On the one hand, `builder()` exists for people that don't want to customize the HTTPS client, but on the other we're asking them to pick the crypto providers (twice!).

I know the upstream rustls crate is moving away from having built-in defaults and I buy that as an argument for why this library crate shouldn't make a choice for the user, but maybe we should remove `builder()` and `builder_with_root()` and make the user provide the `DefaultClient::try_new()` (or `DefaultClient::with_roots()`) result themselves?

It feels a bit clearer to me to have this API take our crypto provider + a HTTPS client vs our crypto provider + a rustls crypto provider.

@djc WDYT?

src/account.rs
```diff
@@ -78,14 +81,18 @@ impl Account {
...
         }
     }
 
-    /// Create an account builder with the given HTTP client
-    pub fn builder_with_http(http: Box<dyn HttpClient>) -> AccountBuilder {
-        AccountBuilder { http }
+    /// Create an account builder with the given [`CryptoProvider`] and HTTP client
+    pub fn builder_with_http(
+        http: Box<dyn HttpClient>,
+        provider: &'static CryptoProvider,
```
cpu, line 98: Consistent with my earlier feedback it feels like `provider` should maybe be the first arg here.

src/account.rs
```diff
@@ -582,61 +593,64 @@ impl AccountBuilder {
...
+            pkcs8,
         ))
     }
 
-    /// Create a new key from the given PKCS#8 DER-encoded private key
-    ///
-    /// Currently, only ECDSA P-256 keys are supported.
-    pub fn from_pkcs8_der(pkcs8_der: PrivatePkcs8KeyDer<'_>) -> Result<Self, Error> {
-        Self::new(pkcs8_der.secret_pkcs8_der(), crypto::SystemRandom::new())
+    pub(crate) fn thumb_sha256(&self) -> Result<[u8; 32], Error> {
```
cpu, line 638: I think `pub(crate)` fns should be ordered below `pub`.

src/account.rs
```diff
@@ -582,61 +593,64 @@ impl AccountBuilder {
...
     }
 
-    fn new(pkcs8_der: &[u8], rng: crypto::SystemRandom) -> Result<Self, Error> {
-        let inner = crypto::p256_key_pair_from_pkcs8(pkcs8_der, &rng)?;
-        let thumb = BASE64_URL_SAFE_NO_PAD.encode(Jwk::new(&inner).thumb_sha256()?);
+    /// Create a key from the given PKCS#8 DER-encoded private key using the given
+    /// [`CryptoProvider`]
+    pub fn from_pkcs8_der(
+        pkcs8_der: PrivatePkcs8KeyDer<'_>,
+        provider: &'static CryptoProvider,
```
cpu, line 649: nit: same feedback RE: reversing the order of args

src/account.rs
```diff
@@ -640,24 +657,26 @@ impl Signer for Key {
 /// See RFC 8555 section 7.3.4 for more information.
 pub struct ExternalAccountKey {
     id: String,
-    key: crypto::hmac::Key,
+    key_value: Vec<u8>,
```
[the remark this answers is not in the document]
cpu: Agree that seems worthwhile.

src/account.rs
```diff
@@ -645,24 +659,26 @@ impl Signer for Key {
...
 }
 
 impl ExternalAccountKey {
-    /// Create a new external account key
+    /// Create a new external account key using the given [`CryptoProvider`]
     ///
     /// Note that the `key_value` argument represents the raw key value, so if the caller holds
     /// an encoded key value (for example, using base64), decode it before passing it in.
-    pub fn new(id: String, key_value: &[u8]) -> Self {
+    pub fn new(id: String, key_value: &[u8], provider: &'static CryptoProvider) -> Self {
```
cpu: same feedback: RE arg order, provider first.

src/account.rs
```diff
@@ -645,24 +659,26 @@ impl Signer for Key {
...
-            key: crypto::hmac::Key::new(crypto::hmac::HMAC_SHA256, key_value),
+            key_value: key_value.to_vec(),
+            hmac_sha256: provider.hmac_sha256,
         }
     }
 }
 
 impl Signer for ExternalAccountKey {
-    type Signature = crypto::hmac::Tag;
+    type Signature = [u8; 32];
```
cpu: flagging this as another place we're choosing to specifically only support HS256. Seems OK, but wanted to call it out.

src/crypto/aws_lc_rs.rs
```diff
@@ -0,0 +1,92 @@
...
+                kty: "EC",
+                x,
+                y,
+            },
+            r#use: "sig",
+        }
+    }
+}
+
+struct BuiltinSha256;
```
cpu: nit: the "Builtin" part of this type name (and the HMAC) one seems a bit odd to me. Could we drop it? (Ditto for the `ring` mod equivalents)

tests/pebble.rs
```diff
@@ -1021,3 +1030,10 @@ impl Drop for Subprocess {
 
 static NEXT_PORT: AtomicU16 = AtomicU16::new(5555);
 const RETRY_POLICY: RetryPolicy = RetryPolicy::new().backoff(1.0);
+
+fn test_provider() -> &'static CryptoProvider {
```
cpu, line 1030: nit: put this above the `static`/`consts`?

src/lib.rs
```diff
@@ -214,7 +216,7 @@ impl DefaultClient {
 
     fn with_roots(
         roots: rustls::RootCertStore,
-        rustls_crypto_provider: CryptoProvider,
+        rustls_crypto_provider: RustlsCryptoProvider,
```
cpu, line 221: same feedback RE: provider as first arg

src/order.rs
```diff
@@ -526,34 +526,45 @@ impl Deref for ChallengeHandle<'_> {
 /// Refer to the methods below to see which encoding to use for your challenge type.
 ///
 /// <https://datatracker.ietf.org/doc/html/rfc8555#section-8.1>
-pub struct KeyAuthorization(String);
+pub struct KeyAuthorization {
+    inner: String,
+    sha256: &'static dyn Sha256,
```
cpu: Would it be feasible/better to pre-compute the SHA256 in `new()` instead of holding on to the `dyn Sha256` ref?

src/types.rs
```diff
@@ -314,6 +314,27 @@ pub enum JwkThumbFields<'a> {
         #[serde(serialize_with = "base64url::serialize")]
         y: &'a [u8],
     },
+    /// Octet Key Pair (Ed25519, Ed448, X25519, etc.)
+    Okp {
+        /// The curve name (e.g., `"Ed25519"`)
+        crv: &'static str,
+        /// Key type, must be `"OKP"`
+        kty: &'static str,
```
cpu: If the field must always be a certain value, should we maybe handle it ourselves in the `Jwk.thumb_sha256` based on `self.alg` or the variant of `self.key`?  I think it means we'd have to have an internal type that has the `kty` field for the purposes of `serde_json`, but it removes one potential source of error from a buggy crypto provider. That seems lightly worth it to me, especially given how many footguns exist in JWS w.r.t "alg: none" and similar :-/

Same feedback applies to the `Rsa.kty` and `Ec.kty` fields if you think it's worth considering.

cpu (commented):
This is really nice work, thank you!

I had a fair amount of feedback but I don't think any of it is super substantial.

src/account.rs
djc: I'd prefer the path to come first (so yes). The provider is sort of ambient; the path is more relevant/particular/specific here.

src/crypto.rs
djc: It doesn't seem to add value given the argument already enforces it, in which case it is just more verbose?

src/account.rs
djc: Yes, but constructors before methods should trump visibility-based ordering.

src/account.rs
cpu: Fair, I think it matches [upstream ordering](https://docs.rs/rustls/latest/rustls/client/struct.WebPkiServerVerifier.html#method.builder_with_provider) in any case. I'll resolve these feedback comments.

src/crypto.rs
cpu: Ok :+1:

src/account.rs
djc: The original deprecation intent was from this comment:

> One minor inconsistency (which probably warrants its own issue) is that this method exposes `PrivateKeyDer` while [AccountBuild::from_parts](https://docs.rs/instant-acme/latest/instant_acme/struct.AccountBuilder.html#method.from_parts) accepts a [PrivatePkcs8KeyDer](https://docs.rs/rustls-pki-types/1.12.0/x86_64-unknown-linux-gnu/rustls_pki_types/struct.PrivatePkcs8KeyDer.html) which is counterintuitive as we may expect the "export side" to give a more specific type (`PrivatePkcs8KeyDer`), and the "import side" to accept, maybe, a more generic type (`PrivateKeyDer`).
> 
> It might be nice to use the same type everywhere and maybe this PR could benefit from changing the type of `AccountCredential.key_pkcs8` (and thus also of this getter) to the more specific `PrivatePkcs8KeyDer`.

In my mind the PKCS#8 is expressed adequately in the type system and so doesn't really merit an additional identifier token.

src/account.rs
djc: Maybe the instant-acme `CryptoProvider` should have a method that returns a `rustls::CryptoProvider`?

Alternatively, agreed that it might make more sense to have callers provide a HTTP client explicitly instead of the complexity here (but in that case we should have the docstring point folks towards `DefaultClient`).

src/types.rs
djc: Yeah, that sounds like an improvement.

src/types.rs
jcgruenhage: Yes.

```diff
diff --git a/src/types.rs b/src/types.rs
index 18dd0d50f7..4494998c71 100644
--- a/src/types.rs
+++ b/src/types.rs
@@ -291,7 +291,7 @@
 }

 impl Jwk<'_> {
-    pub(crate) fn new(key: &crypto::EcdsaKeyPair) -> Jwk<'_> {
+    pub(crate) fn new(key: &crypto::EcdsaKeyPair) -> Self {
         let (x, y) = key.public_key().as_ref()[1..].split_at(32);
         Jwk {
             alg: SigningAlgorithm::Es256,
```

causes 

```
error: lifetime may not live long enough
   --> src/types.rs:296:9
    |
294 |       pub(crate) fn new(key: &crypto::EcdsaKeyPair) -> Self {
    |                              -                         ---- return type is types::Jwk<'2>
    |                              |
    |                              let's call the lifetime of this reference `'1`
295 |           let (x, y) = key.public_key().as_ref()[1..].split_at(32);
296 | /         Jwk {
297 | |             alg: SigningAlgorithm::Es256,
298 | |             key: JwkThumbFields::Ec {
299 | |                 crv: "P-256",
...   |
304 | |             r#use: "sig",
305 | |         }
    | |_________^ associated function was supposed to return data with lifetime `'2` but it is returning data with lifetime `'1`

error: could not compile `instant-acme` (lib) due to 1 previous error
```

An alternative to the current code would be to add lifetime annotations explicitly:

```diff
diff --git a/src/types.rs b/src/types.rs
index 18dd0d50f7..2dd4b49314 100644
--- a/src/types.rs
+++ b/src/types.rs
@@ -290,8 +290,8 @@
     r#use: &'static str,
 }

-impl Jwk<'_> {
-    pub(crate) fn new(key: &crypto::EcdsaKeyPair) -> Jwk<'_> {
+impl<'a> Jwk<'a> {
+    pub(crate) fn new(key: &'a crypto::EcdsaKeyPair) -> Self {
         let (x, y) = key.public_key().as_ref()[1..].split_at(32);
         Jwk {
             alg: SigningAlgorithm::Es256,
```

This also compiles just fine.

jcgruenhage: I've squashed the variant with explicit lifetimes into the commit, will push together with remaining fixes.

src/crypto/aws_lc_rs.rs
jcgruenhage: dropped.

tests/pebble.rs
jcgruenhage: moved.

src/order.rs
jcgruenhage: yes, adjusted accordingly.

src/types.rs
jcgruenhage: I have thought about it a bit and have got a different route: marking the enum + struct variants as non_exhaustive to prevent external construction, and provide functions for construction that set `kty`. This means external construction with wrong `kty` values is not possible anymore.

Is this solution also okay? I can go with an internal type as well, but I'd like to avoid the duplication.

src/crypto.rs
jcgruenhage: Added a short paragraph, does that sound good now?

src/crypto.rs
jcgruenhage: done.

src/crypto.rs
jcgruenhage: This is partially handled, together with https://github.com/djc/instant-acme/pull/268#discussion_r2863763023. The `HmacKey` trait grew a function to get the `SigningAlgorithm`, which is currently lacking variants for HS384 and HS512, but users could still use HS384/HS512 through `SigningAlgorithm::Other`

The one current drawback: A user could construct this with any other `SigningAlgorithm`, like `ES256`, which is not valid

src/account.rs
jcgruenhage: Extended the docs.

src/account.rs
jcgruenhage: Fixed.

src/types.rs
cpu: I haven't looked at the diff yet, but that sounds reasonable as an idea from my perspective :+1:

src/crypto.rs
cpu: Yeah! Looks great, thank you.

src/crypto.rs
cpu: 👍 Thanks

> The one current drawback: A user could construct this with any other SigningAlgorithm, like ES256, which is not valid

I think we might be able to figure out a way to tighten that up too but it doesn't seem immediately pressing or worth blocking this work on.

src/types.rs
cpu: Going to resolve this and djc can unresolve if he would prefer a different solution. Thanks!

cpu:
The latest revisions look good to me. I think the only thing outstanding is the account builder feedback. Do you like the idea of the instant-acme `CryptoProvider` having a method that returns a `rustls::CryptoProvider`? That sounds like an improvement over the status quo.

jcgruenhage:
We're only interacting with `rustls::CryptoProvider` if the `hyper-rustls` feature is active, so I think it'd be better to keep `rustls::CryptoProvider` out of `instant_acme::CryptoProvider`. We could hide it behind a feature flag, but then a any crate exporting a provider would also need to mirror that feature flag. It doesn't feel quite right to me.

jcgruenhage:
Regarding the other suggestion from that thread though, shall I add a change that requires users to pass http client or is that something to consider in the future? I'm happy with either, but given we're already making breaking changes to those functions, I think it makes sense to do it now.

cpu:
> shall I add a change that requires users to pass http client or is that something to consider in the future? 

@djc What do you think? I'm leaning towards saying we should require callers to provide an HTTP client and making it easy for them to construct that, vs trying to make a default HTTP client but still require an explicit ACME cryptography provider. 

In my mind one of the main benefits of the default HTTP client was not needing to muck with the Rustls cryptography provider choice. If we don't want to make a default choice for the ACME crypto provider for the same API experience, I'd be in favor of de-emphasizing a default choice for the HTTP client's cryptography provider and making things consistent.

djc:
> @djc What do you think? I'm leaning towards saying we should require callers to provide an HTTP client and making it easy for them to construct that, vs trying to make a default HTTP client but still require an explicit ACME cryptography provider.

Yeah, I think that makes sense.

> The latest revisions look good to me.

Going to just merge this as is and then tweak some things.

@jcgruenhage thanks for all your work on this!

djc:
> Going to just merge this as is and then tweak some things.

- https://github.com/djc/instant-acme/pull/280

cpu:
> jcgruenhage thanks for all your work on this!

Seconding that, great stuff 👏 Thanks again.

PR #279, opened 2026-03-22
reworked challenge/handle representation, dns-persist-01 support

cpu:
Continuation of https://github.com/djc/instant-acme/pull/265, opened as a new PR since I ended up going a bit of a different direction.

I ended up reworking the challenge state, challenge handle, and key authorization bits fairly substantially before landing the new `dns-persist-01` support. I think the key motivation of the design I landed on is that while it's convenient to try and abstract away authorization challenge response across multiple challenge types, in reality there are a lot of challenge type specific details one has to contend with even when using challenges that all derive the key auth from the token (e.g. the format of the key authorization: should it be the raw string? a SHA256 digest? the base64 of that digest?). Further, the new challenge types (dns-persist-01, device-attest-01) don't have as much in common with the "legacy" challenge types and so we should be using the type system more to differentiate the challenge contexts. This also lets us tighten up what challenges are used for which identifier types (e.g. to avoid a situation where you try to use device-attest-01 with a DNS identifier, or dns-01/dns-persist-01 with an IP address identifier).

Opened as a draft for now:

* I think we could probably get a bit more unit test coverage.
* I haven't tested dns-persist-01 support with any ACME CAs other than Pebble. It would be nice to verify it with Let's Encrypt staging before merge.
* I suspect there's some polish left to do for docs, type names, etc, based on review feedback.

Resolves https://github.com/djc/instant-acme/issues/263

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ specification.
 * Supported extensions:
   * [ACME renewal information (ARI)]
   * [Profiles]
+  * [draft-ietf-acme-device-attest] (**Experimental**)
 * Support for external account binding, key rollover, and contact updates
 * Support for certificate revocation
 * Store/recover your account credentials by serializing/deserializing
@@ -28,6 +29,7 @@ specification.
 
 [ACME renewal information (ARI)]: https://www.rfc-editor.org/rfc/rfc9773.html
 [profiles]: https://datatracker.ietf.org/doc/draft-aaron-acme-profiles/
+[draft-ietf-acme-device-attest]: https://datatracker.ietf.org/doc/draft-ietf-acme-device-attest/
 
 ## Cargo features
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ specification.
   * [ACME renewal information (ARI)]
   * [Profiles]
   * [draft-ietf-acme-device-attest] (**Experimental**)
+  * [draft-ietf-acme-dns-persist] (**Experimental**)
 * Support for external account binding, key rollover, and contact updates
 * Support for certificate revocation
 * Store/recover your account credentials by serializing/deserializing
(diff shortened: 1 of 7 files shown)


cpu:
I'll need to spend some time rebasing this but also wanted to mention that there was a new -01 draft of draft-ietf-acme-dns-persist cut a few days ago and I believe it makes breaking changes relative to the -00 document I used to inform this code. Support for -01 in Pebble hasn't caught up yet either, will keep my eyes on that and update this branch in lockstep with the test server.

djc:
> I'll need to spend some time rebasing this but also wanted to mention that there was a new -01 draft of draft-ietf-acme-dns-persist cut a few days ago and I believe it makes breaking changes relative to the -00 document I used to inform this code. Support for -01 in Pebble hasn't caught up yet either, will keep my eyes on that and update this branch in lockstep with the test server.

After

- #280 

is merged I think it would be good to land the boring parts of this, ideally in a way that inclusion of acme-dns-persist will be a semver-compatible change, so that we can get 0.9 out.

cpu:
Sounds good to me.

cpu:
>  I think it would be good to land the boring parts of this, ideally in a way that inclusion of acme-dns-persist will be a semver-compatible change, so that we can get 0.9 out.

Extracted to https://github.com/djc/instant-acme/pull/283

I'm going to close this for now. I think we should hold on implementing dns-persist-01 until Pebble (and ideally LE staging) are caught up to the latest draft version.

blind-oracle:
JFYI I think Pebble 2.10.1 has caught up mostly: https://github.com/letsencrypt/pebble/pull/544

cpu:
Thanks for the note. I'm hoping to iterate on 283 more soon, and once it's landed I can revive & update the dns-persist-01 code with an up-to-date pebble.

PR #280, opened 2026-03-30
Provider tweaks

djc:
cc @jcgruenhage

PR #283, opened 2026-04-12
Per-type challenge state + handles

cpu:
Extracts the portions of https://github.com/djc/instant-acme/pull/279 that are not specific to `dns-persist-01`.

The key motivation of the design I landed on is that while it's convenient to try and abstract away authorization challenge response across multiple challenge types, in reality there are a lot of challenge type specific details one has to contend with even when using challenges that all derive the key auth from the token (e.g. the format of the key authorization: should it be the raw string? a SHA256 digest? the base64 of that digest?). 

Further, the new challenge types (`dns-persist-01`, `device-attest-01`) don't have as much in common with the "legacy" challenge types and so we should be using the type system more to differentiate the challenge contexts. This also lets us tighten up what challenges are used for which identifier types (e.g. to avoid a situation where you try to use `device-attest-01` with a DNS identifier, or `dns-01`/`dns-persist-01` with an IP address identifier).

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -762,19 +762,89 @@ impl fmt::Display for AuthorizedIdentifier<'_> {
 /// <https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.5>
 #[derive(Debug, Deserialize)]
 pub struct Challenge {
-    /// Type of challenge
-    pub r#type: ChallengeType,
     /// Challenge identifier
     pub url: String,
-    /// Token for this challenge
-    ///
-    /// Unknown `ChallengeType` instances may omit this field, leaving it empty.
-    #[serde(default)]
-    pub token: String,
+    /// Challenge type specific state
+    #[serde(flatten)]
+    pub state: ChallengeState,
     /// Current status
     pub status: ChallengeStatus,
     /// Potential error state
     pub error: Option<Problem>,
+}
+
+/// Challenge type specific state
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum ChallengeState {
+    /// State for an RFC 8555 HTTP-01 challenge
+    Http01(Http01Challenge),
+    /// State for an RFC 8555 DNS-01 challenge
+    Dns01(Dns01Challenge),
+    /// State for an RFC 8737 TLS-ALPN-01 challenge
+    TlsAlpn01(TlsAlpn01Challenge),
+    /// State for a draft-acme-device-attest-08 challenge
+    ///
+    /// Note: Device attestation support is experimental
+    DeviceAttest01,
+    /// An unknown challenge type
+    Unknown(String),
+}
+
+impl ChallengeState {
+    /// Get the challenge type associated with this challenge state
+    pub fn r#type(&self) -> ChallengeType {
+        match self {
+            Self::Http01(_) => ChallengeType::Http01,
+            Self::Dns01(_) => ChallengeType::Dns01,
+            Self::TlsAlpn01(_) => ChallengeType::TlsAlpn01,
+            Self::DeviceAttest01 => ChallengeType::DeviceAttest01,
+            Self::Unknown(r#type) => ChallengeType::Unknown(r#type.clone()),
+        }
+    }
+
+    /// Get the token associated with this challenge (if applicable)
+    ///
+    /// DNS-01, HTTP-01 and TLS-ALPN-01 challenge types offer a token. Other challenge types
+    /// do not rely on RFC 8555 key authorizations and will return `None`, expecting the
+    /// challenge to be satisfied with another method specific to its type.
+    pub fn token(&self) -> Option<&str> {
+        Some(match self {
+            Self::Http01(Http01Challenge { token })
+            | Self::Dns01(Dns01Challenge { token })
+            | Self::TlsAlpn01(TlsAlpn01Challenge { token }) => token,
+            Self::DeviceAttest01 | Self::Unknown { .. } => return None,
+        })
+    }
+}
+
+impl<'de> Deserialize<'de> for ChallengeState {
+    // Deriving `Deserialize` for the `ChallengeState` enum w/ an untagged variant
+    // for unknown challenge types would mean we swallow deser errors for _known_
+    // challenge types. Instead, we want to deser the type string, and then either
+    // capture it as an `Unknown` variant, or dispatch the full value to a more specific
+    // challenge deser, propagating an error result if necessary.
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let value = serde_json::Value::deserialize(deserializer)?;
+
+        let r#type = value["type"]
+            .as_str()
+            .ok_or_else(|| de::Error::missing_field("type"))?;
+
+        match r#type {
+            "http-01" => serde_json::from_value(value)
+                .map(Self::Http01)
+                .map_err(de::Error::custom),
+            "dns-01" => serde_json::from_value(value)
+                .map(Self::Dns01)
+                .map_err(de::Error::custom),
+            "tls-alpn-01" => serde_json::from_value(value)
+                .map(Self::TlsAlpn01)
+                .map_err(de::Error::custom),
+            "device-attest-01" => Ok(Self::DeviceAttest01),
+            unknown_type => Ok(Self::Unknown(unknown_type.to_owned())),
+        }
+    }
 }
 
 /// The challenge type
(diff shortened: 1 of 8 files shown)


djc:
Want to rebase this now that #280 has been merged?

cpu:
> Want to rebase this

Done!

src/types.rs
```diff
@@ -762,21 +762,91 @@ impl fmt::Display for AuthorizedIdentifier<'_> {
...
+}
+
+impl<'de> Deserialize<'de> for ChallengeState {
+    // Deriving `Deserialize` for the `ChallengeState` enum w/ an untagged variant
+    // for unknown challenge types would mean we swallow deser errors for _known_
+    // challenge types. Instead, we want to deser the type string, and then either
+    // capture it as an `Unknown` variant, or dispatch the full value to a more specific
+    // challenge deser, propagating an error result if necessary.
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let value = serde_json::Value::deserialize(deserializer)?;
```
djc: I would prefer to avoid using `Value` here. Have you explored other ways of doing this?

src/types.rs
```diff
@@ -795,6 +865,36 @@ pub enum ChallengeType {
...
+    pub token: String,
+}
+
+/// Challenge state for an RFC 8737 tls-alpn-01 challenge
+///
+/// See <https://www.rfc-editor.org/rfc/rfc8737#section-3>
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
+#[non_exhaustive]
+pub struct TlsAlpn01Challenge {
+    /// A token for constructing a key authorization to complete this challenge
```
djc, line 936: Not totally convinced it's worth having 3 different types for these.

djc (commented):
Here's some early feedback. Need to think more about all the different challenge handling stuff...

src/types.rs
cpu: > Have you explored other ways of doing this?

The best alternative I came up with is using `#[serde(tag = "type")]` + derived deser on the type, with `#[serde(other)]` on the `Unknown` variant.

<details>
<summary>e.g. something like this:</summary>

```rust
#[serde(tag = "type")]
#[non_exhaustive]
pub enum ChallengeState {
    /// State for an RFC 8555 HTTP-01 challenge
    #[serde(rename = "http-01")]
    Http01(Http01Challenge),
    /// State for an RFC 8555 DNS-01 challenge
    #[serde(rename = "dns-01")]
    Dns01(Dns01Challenge),
    /// State for an RFC 8737 TLS-ALPN-01 challenge
    TlsAlpn01(TlsAlpn01Challenge),
    /// State for a draft-acme-device-attest-08 challenge
    ///
    /// Note: Device attestation support is experimental
    #[serde(rename = "device-attest-01")]
    DeviceAttest01,
    /// State for a draft-ietf-acme-dns-persist challenge
    ///
    /// Note: DNS persist challenge support is experimental
    #[serde(rename = "dns-persist-01")]
    DnsPersist01(DnsPersist01Challenge),
    /// An unknown challenge type
    #[serde(other)]
    Unknown,
}
```
</details>

The one "catch" is that `serde(other)` will only work for unit variants and so we lose _what_ the unknown type was. Maybe that's fine? 

Using `#[serde(untagged)]` like you had in https://github.com/djc/instant-acme/pull/265 eats the error from invalid desers of known types. So for example if we had the `dns-persist-01` deser validation we would return `Unknown("dns-persist-01")` for JSON with semantic errors, which both loses the underlying validation failure reason and isn't accurate: the type is known.

I feel pretty strongly we do want the deser validation for the DNS Persist 01 challenge types. There's a lot of MUST constraints in the spec (issuer names list must be 1..10 in len, issuer domain names should be less than 253 octets, not have a trailing dot, A-format, etc) and a parse-don't-verify approach seems best.

If we want to both have sensible deser validation and have a non-unit `Unknown` then I think we need to use `Value` like I have here, or come up with a more complicated `MapAccess` solution with a bunch more code.

TLDR: if you hate `Value`, my recommendation would be that we use a unit type for `Unknown` and accept that we won't express what the unknown challenge type was. WDYT?

src/types.rs
cpu: I tried to motivate this in the PR description but I think it's a worthwhile trade-off to make the API prevent misuse. I think it's a mistake to encourage thinking about all challenges as being equivalent. By specification they aren't ever satisfied in a uniform way and I think the API should reflect that.

As one example from the current API `send_device_attestation()` exists on `Order`, and at runtime produces an error for challenges other than `ChallengeType::DeviceAttest01`. It seems preferable to me that there never be a code path that compiles for trying to use the device attest challenge response submission for an HTTP-01/DNS-01/TLS-ALPN-01/etc challenge.

Similarly, the current API lets you get the BASE64 encoding of the SHA-256 digest of the key authorization of an HTTP-01 challenge token, but that's the incorrect format to use for provisioning a HTTP-01 challenge response. That format _only_ make sense for DNS-01, not HTTP-01 or TLS-ALPN-01. Similar for getting the raw SHA-256 digest of the token without the BASE64 encoding, that's specific to TLS-ALPN-01.

Without separate types we have to offer each of these presentation options and rely on documentation to point out which should be used where, and I think it'd be better to let the type system enforce that.

Does that help convince you? I'm open to other solutions if you think there's a middle ground that can offer the same benefits with fewer moving pieces.

src/types.rs
cpu: >  my recommendation would be that we use a unit type for Unknown and accept that we won't express what the unknown challenge type was. WDYT?

Maybe with HTTP trace logging this isn't a big deal and we can encourage looking at the raw JSON if you need to debug this situation.

cpu: > The best alternative I came up with is using #[serde(tag = "type")] + derived deser on the type, with #[serde(other)] on the Unknown variant.

Switched to that. If you don't like `Value`, it seems like the best solution.

src/types.rs
cpu: @djc Any thoughts here?

src/types.rs
```diff
@@ -762,21 +762,68 @@ impl fmt::Display for AuthorizedIdentifier<'_> {
...
+    #[serde(rename = "device-attest-01")]
+    DeviceAttest01,
+    /// An unknown challenge type
+    #[serde(other)]
+    Unknown,
+}
+
+impl ChallengeState {
+    /// Get the challenge type associated with this challenge state
+    pub fn r#type(&self) -> ChallengeType {
```
djc, line 802: Nit: I think this should be below `token()`.

src/types.rs
djc: Pff, sorry for the long delay.

I'm open to the idea of more strongly anchoring the challenges and their associated state/actions in the type system. I think the per-challenge state types look okay but it feels like there is too much boilerplate in the current iteration of the type-specific challenge handle. I'd also suggest leaving out some of the extra stuff like the `PREFIX` and `webroot_path()` stuff at least for now -- I'm inclined to think users probably don't need help solving that part?

Maybe there should also be a separate `challenges.rs` module with inline modules per challenge type, to make things easier to navigate?

cpu: > I think the per-challenge state types look okay but it feels like there is too much boilerplate in the current iteration of the type-specific challenge handle.

I reworked to remove some of the boilerplate. I think it's improved now.

> I'd also suggest leaving out some of the extra stuff like the PREFIX and webroot_path() stuff at least for now

Removed.

> Maybe there should also be a separate challenges.rs module with inline modules per challenge type, to make things easier to navigate?

Done.

src/types.rs
cpu: Fixed.

cpu:
I gave this branch a fairly substantial rework and I'm happy with where it landed. I think it hits the right trade-offs w.r.t using the type system to guide correct operation while avoiding excessive boilerplate. I also realized we can completely drop the `ChallengeType` enum with this design which is a nice perk relative to how it was before.

A few other things changed along the way (some renaming, avoiding some clones, improving the key auth calculation, dropping earlier tidying commits that were reworked later in the series, etc). I also took the user issue we ran into in discord into account and added some `#[must_use]` annotations to hopefully avoid folks grabbing a handle to a challenge and then doing nothing with it (e.g. forgetting to provision a response and mark the challenge ready).

If you want to see what it would look like to add a new challenge type, I have the original `dns-persist-01` draft support I had started before re-implemented on top here: https://github.com/djc/instant-acme/commit/a604ede878835f415947cb508297f9153616c49c - I didn't include it in this PR because it still needs to be caught up to the latest specification drafts (and there's no rush on that, Let's Encrypt is blocking deployment on https://github.com/ietf-wg-acme/draft-ietf-acme-dns-persist/issues/64 being resolved).

PR #287, opened 2026-04-13
Update dependencies

djc:


cpu (approved):
The README has an outdated MSRV mention that needs updating too. Otherwise lgtm.

cpu:
> CI / msrv (pull_request) Failing after 15s

hmm, might need to change the `time` pin too.

PR #336, opened 2026-08-17
Optional order validity as per RFC 8555 section 7.4

christianhoelzl:
Adds support for requesting an optional certificate validity period when creating an ACME order, as defined in [RFC 8555, Section 7.4](https://www.rfc-editor.org/rfc/rfc8555.html#section-7.4).

- Adds `not_before` and `not_after` to new orders and order states.
- Omits these fields when they are not set.
- Adds unit and Pebble integration tests.

My use case is to control the validity of the requested certificate when the default validity added to the cert by the ACME CA is not a good fit.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -451,6 +451,14 @@ pub struct OrderState {
     /// The profile to be used for the order
     #[serde(default)]
     pub profile: Option<String>,
+    /// The requested value of the notBefore field in the certificate
+    #[cfg(feature = "time")]
+    #[serde(default, with = "time::serde::rfc3339::option")]
+    pub not_before: Option<OffsetDateTime>,
+    /// The requested value of the notAfter field in the certificate
+    #[cfg(feature = "time")]
+    #[serde(default, with = "time::serde::rfc3339::option")]
+    pub not_after: Option<OffsetDateTime>,
 }
 
 /// A wrapper for [`AuthorizationState`] as held in the [`OrderState`]
@@ -496,6 +504,18 @@ pub struct NewOrder<'a> {
     identifiers: &'a [Identifier],
     #[serde(skip_serializing_if = "Option::is_none")]
     profile: Option<&'a str>,
+    #[cfg(feature = "time")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        with = "time::serde::rfc3339::option"
+    )]
+    not_before: Option<OffsetDateTime>,
+    #[cfg(feature = "time")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        with = "time::serde::rfc3339::option"
+    )]
+    not_after: Option<OffsetDateTime>,
 }
 
 impl<'a> NewOrder<'a> {
@@ -507,6 +527,10 @@ impl<'a> NewOrder<'a> {
             identifiers,
             replaces: None,
             profile: None,
+            #[cfg(feature = "time")]
+            not_before: None,
+            #[cfg(feature = "time")]
+            not_after: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


tests/pebble.rs
```diff
@@ -288,6 +288,27 @@ async fn order_deactivate() -> Result<(), Box<dyn StdError>> {
...
+    let env = Environment::new(EnvironmentConfig::default()).await?;
+    let idents = dns_identifiers(["example.com"]);
+    let new_order = &NewOrder::new(&idents)
+        .not_before("3333-01-02T04:00:00Z")
+        .not_after("4444-01-02T04:00:00Z");
+    let mut order = env.account.new_order(new_order).await?;
+
+    // on success the requested validity carries over to pebbles order state.
+    let state = order.state();
+    assert_eq!(state.not_before, Some("3333-01-02T04:00:00Z".into()));
```
cpu: nit: make a const for the date strings passed in to `NewOrder` and asserted against `state`?

src/types.rs
```diff
@@ -451,6 +451,12 @@ pub struct OrderState {
     /// The profile to be used for the order
     #[serde(default)]
     pub profile: Option<String>,
+    /// The requested value of the notBefore field in the certificate (date format is RFC3339)
+    #[serde(default)]
+    pub not_before: Option<String>,
```
cpu: I think we should make this dependent on the optional `time` feature and then implement like:

```rust
  #[cfg(feature = "time")]
  #[serde(default, with = "time::serde::rfc3339::option")]
  pub not_before: Option<OffsetDateTime>,
```

src/types.rs
```diff
@@ -496,6 +502,10 @@ pub struct NewOrder<'a> {
     identifiers: &'a [Identifier],
     #[serde(skip_serializing_if = "Option::is_none")]
     profile: Option<&'a str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    not_before: Option<&'a str>,
```
cpu: and then this would become:

```rust
  #[cfg(feature = "time")]
  #[serde(
      skip_serializing_if = "Option::is_none",
      with = "time::serde::rfc3339::option"
  )]
  not_before: Option<OffsetDateTime>,
```

src/types.rs
```diff
@@ -541,6 +553,18 @@ impl<'a> NewOrder<'a> {
     pub fn identifiers(&self) -> &[Identifier] {
         self.identifiers
     }
+
+    /// Set the requested value of the notBefore field in the certificate (date format is RFC3339)
+    pub fn not_before(mut self, not_before: &'a str) -> Self {
+        self.not_before = Some(not_before);
+        self
+    }
```
cpu: and then this would be something like:

```rust
  #[cfg(feature = "time")]
  pub fn not_before(mut self, not_before: OffsetDateTime) -> Self {
      self.not_before = Some(not_before);
      self
  }
```

cpu (commented):
Thanks for the PR!

I'm generally supportive of this, but I don't think we should make it stringly-typed and lean on docs for the RFC 3339 aspect. Instead I think callers that want to use this feature should have to opt-in to the `time` dependency. WDYT about that idea?

I'm also on the fence about whether some of the rustdoc should explicitly mention that [Let's Encrypt doesn't support this](https://github.com/letsencrypt/boulder/blob/main/docs/acme-divergences.md#section-74). On the one hand we're a general RFC 8555/ACME crate and so should try to avoid over-fitting a specific CA, on the other hand Let's Encrypt is the largest/most popular ACME issuer and I could see users being disappointed when they use this API with them only to get an error. Maybe worth a small note?

christianhoelzl:
I agree that using the specific `OffsetDateTime` type is a better fit for the API than a generic `string`. 

Good point regarding the strict Lets Encrypt validity policy and the documentation change to avoid confusion if the requested validity is not in the enrolled certificate. My point of view is less to enroll from an public ACME CA, but more from private, corporate CAs. There is often more flexibility. For example Step CA does allow to request a valditity within certain limits.

I'll rework the pull request: use `OffsetDateTime`, feature gating, a new description and your other suggestions.

src/types.rs
christianhoelzl: changed

src/types.rs
christianhoelzl: changed

src/types.rs
christianhoelzl: changed

tests/pebble.rs
christianhoelzl: changed

christianhoelzl:
I updated the pull request with the API and documentation changes. The one CI fail is related to a vulnerable h2 dependency version that dependabot should pick up.

cpu (approved):
Thanks, looks great. I opened https://github.com/djc/instant-acme/pull/342 to fix the cargo deny error.

cpu:
> I opened https://github.com/djc/instant-acme/pull/342 to fix the cargo deny error.

Could you rebase on `main`?

christianhoelzl:
is rebased and all CI checks are green.

src/types.rs
```diff
@@ -541,6 +565,30 @@ impl<'a> NewOrder<'a> {
...
+    /// more flexibility over the requested validity.
+    #[cfg(feature = "time")]
+    pub fn not_before(mut self, not_before: OffsetDateTime) -> Self {
+        self.not_before = Some(not_before);
+        self
+    }
+
+    /// Set the requested value of the notAfter field in the certificate
+    ///
+    /// See [`not_before()`][Self::not_before()] for caveats: the ACME server
```
djc: This is a bit of a weird mix between referencing the other method and repeating things anyway? Would be better to pick one and do it.

djc (approved):
Thanks!

src/types.rs
christianhoelzl: I agree, DRY. Removed the second half from the description. I force pushed both commits.

cpu (approved):

PR #348, opened 2026-09-18
Take semver-compatible dependency updates

djc:
In particular, update rustls for [RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285.html).


cpu (approved):
Thanks!

